PR #28, opened 2024-03-22
orb-ui: identical to JJ-prod

fouge:
`orb-ui` as a service, to be installed in next KK release, had many issues that needed to be resolved so that we have the same experience as in JJ-prod, the current version on Orbs.

Hopefully this will address all of the differences.


orb-ui/src/sound/mod.rs
```diff
@@ -314,10 +329,83 @@ impl Player for Jetson {
             tracing::error!("Sound file not found: {:?}", sound_type);
         }
     }
+
+    /// Finds the sound file path for the given sound type and sends it to the
+    /// sound player only if the sink is empty
+    fn try_queue(&self, sound_type: Type) -> Result<bool> {
+        return if let Ok(sink) = self.sink.lock() {
```
TheButlah, line 336: delete the return statement. Rust already returns the last expression.

TheButlah, line 336: instead of doing:
```rust
if let Ok(sink) = self.sink.lock() {
  sink.whatever();
  sink.foo()
} else {
  Err(eyre!("..."))
}
```

use [let .. else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) instead to detect the error and return early, allowing the rest of the code to proceed unindented:

```rust
let Ok(sink) = self.sink.lock() else {
  bail!("...");
};

// Notice how we are no longer indented unecessarily.
sink.whatever();
sink.foo()
```

By using this, I think that basically all of the nested if statments in this function can be un-nested, making it *much* simpler to understand.

orb-ui/src/sound/mod.rs
```diff
@@ -314,10 +329,83 @@ impl Player for Jetson {
...
 }
 
 fn load_filepaths(dir: &str, sound: &str, language: Option<&str>) -> Option<String> {
-    let path = format!("{}/{}{}.wav", dir, sound, language.unwrap_or(""));
+    // if a `language` is passed and the sound is a voice, make sure we append the
+    // localized language to the file name
+    // e.g. voice_server_error__es-ES.wav
+    let has_extension = language.is_some()
+        && !language.unwrap().contains("en-")
+        && sound.contains("voice_");
```
TheButlah, line 397:
```suggestion
    let has_extension = matches!(language, Some(l) if l.contains("en-"))
        && sound.contains("voice_");
```
The [matches macro](https://doc.rust-lang.org/std/macro.matches.html) will return `true` if the left hand argument matches the pattern on the right hand side. Pattern matches in rust support [match guards](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#extra-conditionals-with-match-guards), which allow you to have an `if` statement that uses the pattern matching to further refine the match. 

Combining these allows us to eliminate the `.unwrap()` in favor of explicit pattern matching.
This should be preferred because `.unwrap()` will panic if the code changes over time and the preceding `.is_some()` is removed.

orb-ui/src/sound/mod.rs
```diff
@@ -314,10 +329,83 @@ impl Player for Jetson {
...
+        if let Some(language) = language {
+            format!("__{}", language)
+        } else {
+            "".to_string()
+        }
+    } else {
+        "".to_string()
+    };
+
+    let path = format!("{}/{}{}.wav", dir, sound, lang_extension);
```
TheButlah, line 408: generally, you should be using `std::path::Path` and `.join()` instead of manually formatting your filepaths as strings.

orb-ui/src/sound/mod.rs
```diff
@@ -408,13 +496,59 @@ impl Player for Fake {
...
+                            sound_type
+                        );
+                    }
+                } else {
+                    tracing::error!("Sound file not found: {:?}", sound_type);
+                }
+            }
+        }
+        Ok(false)
+    }
```
TheButlah, line 525: same thing here, almost all of this can be collapsed such that its not a bunch of nested ifs.

orb-ui/src/sound/mod.rs
```diff
@@ -408,13 +496,59 @@ impl Player for Fake {
...
+
+    fn set_volume(&self, volume_percent: u64) {
+        if let Ok(sink) = self.sink.lock() {
+            sink.set_volume((volume_percent as f64 / 100_f64) as f32);
+        } else {
+            tracing::error!("Failed to lock sink to set sound volume");
```
TheButlah, line 531: This failure mode will go away when we get rid of the mutex, but for future reference:

If the mutex fails to lock, it means it is poisoned and something very wrong happened. Logging is probably not the right behavior, I would panic.

TheButlah (changes requested):

orb-ui/src/sound/mod.rs
```diff
@@ -352,19 +441,18 @@ impl Fake {
         // Get a output stream handle to the default physical sound device
         let (stream_task, stream_handle) =
             StreamTask::new().wrap_err("failed to create stream task")?;
-        let sink = rodio::Sink::try_new(&stream_handle)?;
-        let (tx, mut rx) = mpsc::channel(1);
-
-        // TODO load config
-
+        let sink = Arc::new(Mutex::new(rodio::Sink::try_new(&stream_handle)?));
+        let (tx, mut rx) = mpsc::channel(5);
```
TheButlah, line 445: magic number alert: I suggest making it a const and putting it at the top of the file. Also, why 5 and not 1 or 0, or a power of 2?

orb-ui/src/engine/mod.rs
```diff
@@ -70,9 +70,6 @@ macro_rules! event_enum {
                 $(#[doc = $doc])?
                 fn $method(&self, $($($field: $ty,)*)?);
             )*
-
-            /// Returns a new handler to the shared queue.
```
TheButlah, line 74: It seems to me that it is convenient for users to be able to clone the engine. You have presented a synchronized "proxy" object to acces the sound, why not allow people to clone it so that they can pass it to various tasks/threads?

orb-ui/src/sound/mod.rs
fouge: now using PathBuf, which is cross platform, instead of these hardcoded filepaths

orb-ui/src/engine/mod.rs
fouge: I just thought I would remove unused code

fouge:
thank you @TheButlah for the review 👍 
I took the comments into account :) 

I don't have time for the optional stuff though 😁

orb-ui/examples/ui-replay.rs
```diff
@@ -47,37 +79,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let _ = proxy
-        .orb_signup_state_event(format!("\"SoundVolume {{ level: 10 }}\""))
+        .orb_signup_state_event("\"SoundVolume {{ level: 10 }}\"".to_string())
         .await;
 
     // get path to records file from program arguments or use default
-    let path = std::env::args().nth(1).unwrap_or("records.txt".to_string());
-    let file = File::open(path)?;
+    let path = args.path.unwrap_or(RECORDS_FILE.to_string());
+    let file = File::open(path.clone()).wrap_err(format!("cannot open {path}"))?;
```
TheButlah, line 95: nit: if the error type is expensive to create (such as allocating a string), then you want to lazily construct it. `wrap_err` constructs it eagerly, whereas `wrap_err_with` does this lazily with a closure that gets invoked *only* on the error case.
```suggestion
    let file = File::open(path.clone()).wrap_err_with(|| format!("cannot open {path}"))?;
```

orb-ui/src/sound/capture.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl Default for CaptureLoopSound {
+    fn default() -> Self {
+        CaptureLoopSound::Loop01(LoopSoundPart::PartA)
+    }
+}
+
+impl CaptureLoopSound {
+    /// Come back to part A when out of range detected
+    pub fn out_of_range(&mut self) {
```
TheButlah, line 28: the name of this function makes it sound like this detects when it is out of range, rather than "do thing when we are out of range".

Lets think of a more descriptive name, Maybe, restart_loop?

orb-ui/src/sound/mod.rs
```diff
@@ -476,60 +463,43 @@ impl Player for Fake {
...
-                }
-            } else {
-                tracing::error!(
-                    "Sound file {:?} doesn't have a known file path",
-                    sound_type
-                );
-            }
+    fn queue(&self, sound_type: Type) -> Result<()> {
+        let Some(sound_file) = self.sound_files.get(&sound_type) else {
+            eyre::bail!("Sound not found: {:?}", sound_type);
```
TheButlah, line 472: nit: `use eyre::bail;`

orb-ui/src/sound/mod.rs
```diff
@@ -238,8 +252,19 @@ sound_enum! {
     }
 }
 
+/// Default sound volume
+const DEFAULT_SOUND_VOLUME_PERCENT: u64 = 10;
+
+pub struct Jetson {
+    _stream_task: StreamTask,
+    _stream_handle: rodio::OutputStreamHandle,
+    queue_file: Mutex<mpsc::Sender<PathBuf>>,
```
TheButlah, line 261: a `mpsc` implements `Sync` and also can be cloned. You should be able to just get a &mut to it fine without a mutex. However, your other methods on `Jetson` will need to take `&mut Self` 

Thats fine, because the jetson is designed to be a cloneable, `Sync` proxy object. So you can be sure that you are gonna have exclusive access to self - if it needed to be shared, the user would just clone it.

orb-ui/src/sound/mod.rs
```diff
@@ -388,39 +463,68 @@ impl Player for Fake {
...
+            eyre::bail!("Sound {:?} doesn't have a known file path", sound_type);
+        };
+
+        if let Ok(mut tx_queue) = self.queue_file.lock() {
+            tx_queue
+                .try_send(sound_file.clone())
+                .wrap_err("Failed to queue sound")?;
         } else {
-            tracing::error!("Sound file not found: {:?}", sound_type);
+            eyre::bail!("Failed to lock queue")
```
TheButlah, line 484: instead of logging, or returning Err, we should panic if the lock is poisoned.

orb-ui/src/sound/mod.rs
```diff
@@ -388,39 +463,68 @@ impl Player for Fake {
...
+
+        self.queue(sound_type).map(|_| true)
+    }
+
+    fn set_volume(&self, volume_percent: u64) {
+        self.sink
+            .set_volume((volume_percent as f64 / 100_f64) as f32);
+    }
+
+    fn set_language(&self, language: Option<&str>) {
```
TheButlah, line 505: Shouldn't this return a Result?

TheButlah (changes requested):

orb-ui/src/sound/capture.rs
```diff
@@ -24,8 +24,9 @@ impl Default for CaptureLoopSound {
 }
 
 impl CaptureLoopSound {
-    /// Come back to part A when out of range detected
-    pub fn out_of_range(&mut self) {
+    /// Come back to part A of the current playing loop sound
+    /// (when out of range detected)
```
TheButlah, line 28:
```suggestion
```
This function unconditionally restarts the current loop. There is no check for out of range from what I could see. Can you confirm?

TheButlah (approved):
other than the question I had about the loop, this looks good.

TheButlah (approved):

PR #40, opened 2024-04-03
orb-ui: custom sound crate and include .service in deb

fouge:
taken from orb-core
uses _official_ `alsa-sys` crate instead of custom `alsa-sys` crate

now excludes orb-ui from macos tests (not cross platform anymore with this PR)


orb-ui/debian/orb-ui.worldcoin-ui.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Worldcoin UI Service
+Requires=worldcoin-dbus.service
```
oldgalileo, line 3: Better to depend on the `.socket` (`worldcoin-dbus.socket`)

orb-ui/debian/orb-ui.worldcoin-ui.service
```diff
@@ -0,0 +1,20 @@
...
+Requires=dbus.socket
```
oldgalileo, line 4: Why do you need the system bus at all?

orb-ui/debian/orb-ui.worldcoin-ui.service
```diff
@@ -0,0 +1,20 @@
...
+Environment="DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/worldcoin_bus_socket"
+Environment="RUST_BACKTRACE=1"
+SyslogIdentifier=worldcoin-ui
+WorkingDirectory=/home/worldcoin
+ExecStart=/usr/local/bin/orb-ui daemon
+Restart=always
+
+[Install]
+; TODO: maybe wanted by other things too
+WantedBy=worldcoin-backend-connect.target
```
oldgalileo, line 20: This shouldn't be the case. This dependency states that for `worldcoin-backend-connect.target` to succeed (and for things which `Require` it to continue), that the `worldcoin-ui` service must be started.

This should likely be completely orthogonal to the networking. When you install this to the `WantedBy=multi-user.target` and then within the `worldcoin-backend-connect.service` add `After=worldcoin-ui.service` and `Wants=worldcoin-ui.service`, this will effectively startup the `worldcoin-backend-connect.service` after `worldcoin-ui` has been kicked off. Importantly, this will not pull down the whole networking bringup in the event that there's a UI issue.

oldgalileo (changes requested):

oldgalileo (approved):
Only approving the `.service` file. Have not looked through the source (cc @andronat )

TheButlah (approved):

TheButlah:
Approving. I didn't review the sound crate because it was pulled straight from JJ orb-core.

PR #43, opened 2024-04-04
Install orb-attest and orb-ui to /usr/local/bin

fouge:
Previously, we had these binaries installing to the default location of `/usr/bin`. However, most of the custom binaries we include in Orb OS are installed to `/usr/local/bin`. Updating for consistency.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -49,7 +49,7 @@ jobs:
     steps:
       - name: Ensure git ref is valid for the selected channel
         env:
-          CI_DEFAULT_BRANCH=${{ github.event.repository.default_branch }}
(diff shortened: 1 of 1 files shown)


TheButlah (changes requested):
This adds additional complexity and brittleness to the .deb packaging metadata, but more importantly cargo-deb doesn't support changing the `release` part of the path based on the target profile (can be `release` or `artifact`). 

I think we should not merge this change, given that having things in /usr/local/bin is merely stylistic preference, and also other binaries were already in /usr/bin (orb-short-lived-token-daemon, slot-ctrl)

TheButlah:
Sorry I accidentally force pushed which closed the PR. Feel free to reopen if its necessary to further discuss my concern with the /usr/bin issue.

PR #68, opened 2024-04-05
orb-ui: resurrect network connection success sound

dangirsh:
used to indicate that wifi has successfully connected

only played after wifi qr is requested

goes with https://github.com/worldcoin/priv-orb-core/pull/1141


karelispanagiotis:
Tested that both changes in `orb-core` and `orb-ui` work!

karelispanagiotis (approved):
LGTM

PR #74, opened 2024-04-15
orb-ui: Correctly clear ring animations on SignupFail

valff:


karelispanagiotis (commented):
This PR solves both UI glitches:
1. Clear LED ring on unsuccessful signup
2. LED ring slider arc remains on Biometric Capture timeout 

After this change, the Biometric Capture Timeout flashes the LED ring circle **twice**, but it shouldn't. Can we also fix that @valff ?

PR #79, opened 2024-04-17
ci: make MAC_EXCLUDE a set of inputs

TheButlah:


TheButlah:
This is not necessary due to #104

PR #84, opened 2024-04-30
orb-ui: remove logging of each UI event

fouge:
too loud in the logs


TheButlah (approved):
approved, but I edited the PR to downgrade it to trace level instead of fully removing.

TheButlah (approved):

PR #85, opened 2024-05-02
mcu-util: blocking call to can-rs in async 

fouge:
and some tiny refacto

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,6 +11,7 @@ on:
           - orb-attest
           - orb-thermal-cam-ctrl
           - orb-ui
+          - orb-mcu-util
       channel:
         description: |
           Which release channel? 
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -116,13 +116,13 @@ jobs:
         if: ${{ matrix.platform == 'macos-13' }}
         run: |
           MAC_EXCLUDE=(
+            "--exclude can-rs"
+            "--exclude orb-mcu-util"
+            "--exclude orb-sound"
             "--exclude orb-thermal-cam-ctrl"
+            "--exclude orb-ui"
             "--exclude seek-camera"
             "--exclude seek-camera-sys"
-            "--exclude can-rs"
-            "--exclude mcu-util"
-            "--exclude orb-sound"
-            "--exclude orb-ui"
           )
           echo MAC_EXCLUDE="${MAC_EXCLUDE[*]}" >>${GITHUB_ENV}
           cat ${GITHUB_ENV}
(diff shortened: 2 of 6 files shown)


mcu-util/src/messaging/can/canfd.rs
```diff
@@ -201,26 +201,20 @@ impl MessagingInterface for CanRawMessaging {
         };
 
         if let Some(bytes) = bytes {
-            let mut buf: [u8; 64] = [0u8; 64];
+            let mut buf: [u8; CANFD_DATA_LEN] = [0u8; CANFD_DATA_LEN];
```
TheButlah, line 204:
```suggestion
            let mut buf = [0u8; CANFD_DATA_LEN];
```
No need to annotate, the type is already clear.

mcu-util/src/messaging/can/canfd.rs
```diff
@@ -201,26 +201,20 @@ impl MessagingInterface for CanRawMessaging {
...
             buf[..bytes.len()].copy_from_slice(bytes.as_slice());
 
             let node_addr = self.can_node as u32;
             let frame = Frame {
                 id: Id::Extended(node_addr),
-                len: 64,
+                len: CANFD_DATA_LEN as u8,
```
TheButlah, line 210: May be a typo: I thought it would be bytes.len()?
```suggestion
                len: bytes.len(),
```

mcu-util/src/messaging/can/isotp.rs
```diff
@@ -256,6 +260,6 @@ impl MessagingInterface for CanIsoTpMessaging {
             _ => return Err(eyre!("Invalid payload")),
         };
 
-        self.send_wait_ack(bytes.as_slice()).await
+        self.send_wait_ack(Arc::new(bytes)).await
```
TheButlah, line 263: fyi: the [bytes](https://docs.rs/bytes) crate is typically used for reference counted byte buffers, and you'll encounter it throughout much of the tokio ecosystem. There is nothing wrong with using a `Arc<Vec<u8>>` for simple use cases like this though - *if* you need shared references.

However, I don't see why an Arc is needed. Can't you just send the Vec directly through the channel without wrapping it in an Arc?

TheButlah (changes requested):

mcu-util/src/messaging/can/isotp.rs
fouge: the issue is within `send_wait_ack`, I need to pass the `Vec` to another thread

mcu-util/src/messaging/can/canfd.rs
fouge: fixed

TheButlah: please confirm that this change is correct - I was guessing, and wanted to defer to you on the decision on whether it would be CANFD_DATA_LEN or bytes.len().

mcu-util/src/messaging/can/isotp.rs
TheButlah: > I need to pass the Vec to another thread
Thats fine: by making `send_wait_ack` take ownership of the vec, you can subsequently send the vec to the other thread. No copy required.

PR #89, opened 2024-05-03
wpa-supplicant: added crate

TheButlah:
Pulled out of https://github.com/worldcoin/priv-orb-core/pull/841, so we don't lose the work and since this crate doesn't have any dependencies on orb-core.

I also took the opportunity to upgrade to zbus 4 and added (very small) bit of documentation.


Cargo.toml
```diff
@@ -31,19 +32,23 @@ repository = "https://github.com/worldcoin/orb-software"
 rust-version = "1.77.0" # See rust-version.toml
 
 [workspace.dependencies]
+alsa-sys = "0.3.1"
```
fouge, line 35: `alsa-sys` is used only by the orb-ui/sound crate, and won't be used by another crate I think, we might want to move it to the Cargo.toml of that crate, what do you think? 🤔

TheButlah: Sure. I don't actually have a strong opinion either way. We can do that.

TheButlah:
Its been a while since review was requested from @oldgalileo, but we never got one. Forcing a merge according to the team policy.

PR #90, opened 2024-05-04
can-rs: explicitly check number of received can bytes

TheButlah:


can/src/stream.rs
```diff
@@ -416,21 +416,29 @@ mod imp {
...
+                std::mem::size_of::<RawFrame<{ N }>>(),
                 flags,
                 addr,
                 addrlen,
             )
         };
-        if ret < 0 {
+        if nbytes < 0 {
             return Err(io::Error::last_os_error());
+        } else if nbytes as usize != std::mem::size_of::<RawFrame<{ N }>>() {
```
oldgalileo, line 433: This does not hold up, as when `N = 64` the acceptable values of `nbytes` are either `size_of::<RawFrame<N=8>>` OR `size_of::<RawFrame<N=64>>`

oldgalileo (changes requested):

fouge:
@TheButlah yeah i don't think this PR makes sense.
The frame length isn't 8 or 64 (CAN-FD), it can take a few different values, see the [DLC field](https://elearning.vector.com/mod/page/view.php?id=368)

I'm closing the PR

PR #94, opened 2024-05-07
cargo: centralize features into workspace

TheButlah:
Currently, CI only builds with `--all` and doesn't go build individual crates. This means that cargo features are [unified across the workspace](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification).

There are two footguns with this:
* Building a crate directly with -p may reduce the set of features in use to exactly what is in that crate`s cargo.toml. This can break the build if we forgot to specify all the needed features, and we don't check that in CI.
* Changing the features on a crate likely causes the crate to need to be rebuilt. This means that all dependent crates also need to be rebuilt.

To improve both those footguns, it makes sense to just specify features on common crates directly in the workspace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1905,7 +1905,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.4.10",
+ "socket2 0.5.5",
  "tokio",
  "tower-service",
  "tracing",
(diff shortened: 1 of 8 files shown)


fouge (approved):
lgtm
whenever one feature has to be enabled for a package but not another, should we specify the dependency at the package level and not the workspace level?

TheButlah:
> should we specify the dependency at the package level and not the workspace level?

Cargo will unify features across the workspace, regardless of whether the crates were specified in the workspace dependencies or crate dependencies.

 I don't think there would be any harm in pulling in more features than we strictly need to since all features are supposed to be additive. So I would just put those features at the workspace level, yeah.

PR #95, opened 2024-05-07
build-info: add orb prefix and improve caching

TheButlah:
* fixes a bug in build-info that caused it to always rebuild.
* makes orb-slot-ctrl use build-info.
* renames the crate to orb-build-info

### DIFF
--- a/orb-slot-ctrl/src/main.rs
+++ b/orb-slot-ctrl/src/main.rs
@@ -1,6 +1,10 @@
 use clap::{Parser, Subcommand};
 use std::{env, process::exit};
 
+use orb_build_info::{make_build_info, BuildInfo};
+
+const BUILD_INFO: BuildInfo = make_build_info!();
+
 #[derive(Parser)]
 #[command(
     author,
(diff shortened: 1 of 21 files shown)


orb-slot-ctrl/src/main.rs
```diff
@@ -192,8 +196,8 @@ fn main() -> eyre::Result<()> {
                 }
             }
         }
-        Commands::GitCommit => {
-            println!("{}", env!("GIT_COMMIT"));
+        Commands::GitDescribe => {
+            println!("{}", BUILD_INFO.git.describe);
```
TheButlah, line 200: Pointing this out: AFAIK no one relies on this command, please correct me if I am wrong.

fouge: I personally have no idea 🤔

fouge (approved):
I am not the best person to review this but lgtm
I disabled the auto-merge feature so that maybe someone else can have a look. 

Some conflicts have to be resolved as well

PR #97, opened 2024-05-08
hil: initial commit

TheButlah:
Adds an orb-hil binary, to make flashing easy and automated. Its little more than a glorified bash script right now.

Currently it:
* Assumes that it is on the hil and relies on some cli tools provided from the nix shell, like awscli and tar
* Supports flashing from a s3 url or a local tarball

It does not yet support:
* Automatically controlling CTS and RTS pins to reboot orb or trigger recovery mode
* Connecting to serial
* Retrieving s3 urls from a git tag
* Automatically running in response to network events or github runners


hil/src/flash.rs
```diff
@@ -0,0 +1,82 @@
...
+    ensure!(
+        path_to_rts.try_exists().unwrap_or(false),
+        "{path_to_rts} doesn't exist"
+    );
+    ensure!(path_to_rts.is_file(), "{path_to_rts} should be a file!");
+    let temp_dir = TempDir::new_in(path_to_rts.parent().unwrap())
+        .wrap_err("failed to create temporary extract dir")?;
+    let extract_dir = temp_dir.path();
+    let result = run_cmd! {
+        cd $extract_dir;
```
alekseifedotov, line 48: these bash snippets scare me.

TheButlah: its not actually bash. This is just a macro that calls std::process:Command. You can read more here: https://docs.rs/cmd_lib/latest/cmd_lib/#why-you-need-this

alekseifedotov (approved):

PR #98, opened 2024-05-13
security-utils: minor improvements

TheButlah:


security-utils/src/reqwest.rs
```diff
@@ -7,6 +7,8 @@ use color_eyre::{
 use hex_literal::hex;
 use reqwest::{Certificate, Client, ClientBuilder};
 
+pub use reqwest;
+
```
fouge, line 11: should this line be part of the PR?

TheButlah: Yes, it re-exports the reqwest crate. This can be useful to use the same crate version across dependencies.

fouge (approved):

PR #101, opened 2024-05-17
ci: autodetect cargo deb crates

TheButlah:
Lets us stop having to manually write out the list of binaries. Also will allow us to reuse this workflow in orb-internal, since it is no longer specific to this repo.

Also seeking opinions: Would it be better for me to keep the json parsing in jq, or do you want me to port it to native python?

### DIFF
--- a/scripts/build_rust_artifacts.py
+++ b/scripts/build_rust_artifacts.py
@@ -7,52 +7,87 @@ import os
 import shlex
 
 
-def cmd(command):
+def run(command):
     assert isinstance(command, str)
-    # This allows us to avoid manually splitting strings into arguments
-    command = shlex.split(command)
-    print(f"Running: {' '.join(command)}")
-    subprocess.check_call(command)
+    print(f"Running: {command}")
+    exit_code = subprocess.check_call(command, shell=True, text=True)
+    assert exit_code == 0
 
 
-def main():
-    parser = argparse.ArgumentParser(description="Builds rust artifacts for CI")
-    parser.add_argument("out_dir", help="Output directory for artifacts")
-    parser.add_argument(
-        "cargo_profile", help="Cargo profile to use for compiling the crates"
+def run_with_stdout(command):
+    assert isinstance(command, str)
+    print(f"Running: {command}")
+    cmd_output = subprocess.check_output(command, shell=True, text=True)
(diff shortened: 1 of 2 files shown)


scripts/build_rust_artifacts.py
```diff
@@ -7,53 +7,88 @@
...
+    assert isinstance(command, str)
+    print(f"Running: {command}")
+    cmd_output = subprocess.check_output(command, shell=True, text=True)
+    return cmd_output
 
-    args = parser.parse_args()
 
-    targets = ["aarch64", "x86_64"]
+def find_cargo_deb_crates():
+    jq_query = (
```
andronat, line 25: The alternative in Python would be (something along these lines):
```
    metadata = json.loads(cmd_output)
    workspace_members = set(metadata['workspace_members'])
    crates = [
        package['name']
        for package in metadata['packages']
        if package['id'] in workspace_members and 'deb' in package.get('metadata', {})
    ]
```

I personally find it more understandable. But I'll leave this to you. 

Is `jq` always available btw? Because with pure python implementation you have the advantage of "peace of mind".

scripts/build_rust_artifacts.py
```diff
@@ -7,53 +7,88 @@
...
                 f"cp -L "
-                f"target/{t}-unknown-linux-gnu/{args.cargo_profile}/{b} "
-                f"{args.out_dir}/{b}/{b}_{t}"
+                f"target/{t}-unknown-linux-gnu/{cargo_profile}/{c} "
+                f"{out_dir}/{c}/{c}_{t}"
             )
 
 
+def main():
+    parser = argparse.ArgumentParser(description="Builds rust artifacts for CI")
```
andronat, line 69: very nit for whenever (if) you have time: It would be amazing to have a description that our artifacts are actually debian packages, and also have the description declaring what are the assumptions under which we can build a debian.

andronat (approved):

scripts/build_rust_artifacts.py
TheButlah: we ensure that jq is always available via the nix flake. But yes, I think I prefer the python approach more. Also, I know python more than I know jq. I'll open a new PR to port it.

PR #103, opened 2024-05-17
ci: generalize MAC_EXCLUDE

TheButlah:
In order to reuse rust-ci.yaml, we need to automatically exclude crates that don't work on various platforms, without manually specifying them in CI.

To do this, I first introduced a custom metadata section for first party crates under [package.metadata.orb]. This can be used going forward as a place to put custom metadata for our crates that cargo subcommands and CI can inspect.

Next, I added a custom field under it called `unsupported_targets`, which lists out any of the target triples that CI runs that the crate is known to be incompatible with. 

Finally, I created a script to parse this data with jq and then call it in CI to get the crate exclusions for the current host architecture.

### DIFF
--- a/scripts/get_excludes.py
+++ b/scripts/get_excludes.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+
+
+import argparse
+import subprocess
+import sys
+import os
+import shlex
+
+
+def run_with_stdout(command):
+    assert isinstance(command, str)
+    print(f"Running: {command}", file=sys.stderr)
+    cmd_output = subprocess.check_output(command, shell=True, text=True)
+    return cmd_output
+
+
+def get_target_triple():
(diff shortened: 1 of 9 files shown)


scripts/get_excludes.py
```diff
@@ -0,0 +1,45 @@
...
+    for s in cmd_output:
+        if s.startswith("host:"):
+            return s.split(" ")[1]
+    raise Exception("no target triple detected")
+
+
+def main():
+    target = get_target_triple()
+    print(f"identified target triple as: {target}", file=sys.stderr)
+    jq_query = (
```
andronat, line 29: just for visibility ditto with: https://github.com/worldcoin/orb-software/pull/101/files#r1605355834

andronat (approved):

PR #104, opened 2024-05-21
ci: use subcommands in helper and stop using jq

TheButlah:
In #101 and #103, I agreed to port the code from jq to pure python. This pr implements that change and also merges the two helper python scripts into one file.

I also fixed a regression where not all cargo debs were being produced, and not all binaries were getting copied into the artifacts dir.


scripts/rust_ci_helper.py
```diff
@@ -0,0 +1,146 @@
...
+import json
+import argparse
+import subprocess
+import os
+
+
+def run(command):
+    assert isinstance(command, str)
+    print(f"Running: {command}")
+    exit_code = subprocess.check_call(command, shell=True, text=True)
```
andronat, line 12: Even if you don't use the output (both stdout/err) from the command, I would suggest it's still useful to print it in case something goes wrong.

scripts/rust_ci_helper.py
```diff
@@ -0,0 +1,146 @@
...
+    assert exit_code == 0
+
+
+def run_with_stdout(command, print_cmd=True):
+    assert isinstance(command, str)
```
andronat, line 17: That's kinda a useless check. very broad.

scripts/rust_ci_helper.py
```diff
@@ -0,0 +1,146 @@
...
+    if print_cmd:
```
andronat, line 18: I would say just always print it. you are writing some CI middlewere here. visibility is important.

scripts/rust_ci_helper.py
TheButlah: I see it strictly being beneficial for error detection. Can you elaborate a bit why you don't like it?

scripts/rust_ci_helper.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def run(command):
+    assert isinstance(command, str)
+    print(f"Running: {command}")
+    exit_code = subprocess.check_call(command, shell=True, text=True)
+    assert exit_code == 0
+
+
+def run_with_stdout(command, print_cmd=True):
```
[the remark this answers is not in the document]
TheButlah: The reason why run_with_stdout and run both exist is that the latter *captures* stdout and wouldn't print it until execution is finished. Whereas the former just captures the error code, and therefore stdout is streamed in realtime.

I *could* create a way to unify these two such that stdout is always printed in realtime, while still letting me capture the output as a python string, but that is more work than I think is worth.

I've pushed a change that *always* prints though, and removes the `print_cmd` flag.

TheButlah:
Side note: writing this script really made me regret not just using rust. Having an actual type checker so I didn't have to guess all the time, would have been really nice.

andronat (approved):

PR #110, opened 2024-05-28
hil: detect recovery mode

TheButlah:


TheButlah:
I will force a merge on this by EOD if no review comes.

PR #118, opened 2024-06-04
orb-mcu-util: don't fail on build

fouge:
when fetching data, don't fail
but prefer unfilled struct

@TheButlah i'd like your thought on this kind of construct. I think it's fine not to fail when fetching the data so that we can at least print the pieces of info that were fetched.

### DIFF
--- a/mcu-util/src/orb/main_board.rs
+++ b/mcu-util/src/orb/main_board.rs
@@ -105,11 +105,14 @@ impl Board for MainBoard {
     }
 
     async fn fetch_info(&mut self, info: &mut OrbInfo) -> Result<()> {
-        let main = MainBoardInfo::new().build(self).await;
+        let board_info = MainBoardInfo::new()
+            .build(self)
+            .await
+            .unwrap_or_else(|board_info| board_info);
 
-        info.hw_rev = main.hw_version;
-        info.main_fw_versions = main.fw_versions;
-        info.main_battery_status = main.battery_status;
+        info.hw_rev = board_info.hw_version;
+        info.main_fw_versions = board_info.fw_versions;
+        info.main_battery_status = board_info.battery_status;
 
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


mcu-util/src/orb/main_board.rs
```diff
@@ -328,9 +328,10 @@ impl MainBoardInfo {
     }
 
     /// Fetches `MainBoardInfo` from the main board
+    /// doesn't fail, but lazily fetches as much info as it could
     /// on timeout, returns the info that was fetched so far
-    async fn build(mut self, main_board: &mut MainBoard) -> Result<Self> {
-        main_board
+    async fn build(mut self, main_board: &mut MainBoard) -> Self {
```
TheButlah, line 333: Instead, make this a `Result<Self, Self>`. The err variant is when not all data was fetched.

mcu-util/src/orb/main_board.rs
```diff
@@ -328,9 +328,10 @@ impl MainBoardInfo {
     }
 
     /// Fetches `MainBoardInfo` from the main board
+    /// doesn't fail, but lazily fetches as much info as it could
     /// on timeout, returns the info that was fetched so far
```
TheButlah, line 332: At first, I was about to comment asking why would we have a timeout in async code since the caller can just use `tokio::time::timeout()`. Then I investigated further, and noticed that we call blocking code from this async function. I thought we removed all of that 😭

I'll open a separate PR to fix this, since its unrelated to this PR. It is *extremely important* not to block in async code, as it leads to very hard to debug behavior and is often a cause for subtle bugs.

mcu-util/src/orb/security_board.rs
```diff
@@ -340,8 +340,8 @@ impl SecurityBoardInfo {
 
     /// Fetches `SecurityBoardInfo` from the security board
     /// on timeout, returns the info that was fetched so far
-    async fn build(mut self, sec_board: &mut SecurityBoard) -> Result<Self> {
-        sec_board
+    async fn build(mut self, sec_board: &mut SecurityBoard) -> Self {
```
TheButlah, line 343: Same here as in the comments on the main board.

TheButlah (changes requested):
I like the general approach, left some requests to be more explicit about the partial initialization.

mcu-util/src/orb/main_board.rs
TheButlah: https://linear.app/worldcoin/issue/O-2706/blocking-in-async-in-main-board-and-sec-board

TheButlah (approved):

PR #119, opened 2024-06-04
docs: Make SEEK_SDK_OVERRIDE more consistent across documentation and…

AlexKaravaev:
When setting up repo https://github.com/worldcoin/orb-software/blob/main/CONTRIBUTING.md tells you to set SEEK_SDK_OVERRIDE in .envrc.example in which VENDORED_SEEK_PATH is used.

Also seek-camera will ignore both VENDORED_SEEK_PATH/SEEK_SDK_OVERRIDE. 

Not a big change but improves onboarding process a bit.

### DIFF
--- a/.envrc.example
+++ b/.envrc.example
@@ -1,5 +1,5 @@
-# VENDORED_SEEK_PATH=~/P/wc/seek-thermal-sdk
-# use flake --override-input seekSdk $VENDORED_SEEK_PATH
+# FLAKE_OVERRIDE_SEEK_SDK=~/P/wc/seek-thermal-sdk
+# use flake --override-input seekSdk $FLAKE_OVERRIDE_SEEK_SDK
 use flake
 
 # export CACHIX_AUTH_TOKEN=<TOKEN>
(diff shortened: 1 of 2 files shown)


seek-camera/sys/build.rs
```diff
@@ -60,7 +60,11 @@ fn main() -> Result<()> {
         // We will allow users to specify the SDK path if they don't want to have it system installed.
         // For example, if on mac, you can use this to get around the fact that the seek sdk only
         // works on windows and linux.
-        let sdk_path = env::var("SEEK_SDK_PATH").ok().map(PathBuf::from);
+        let sdk_path = env::var("SEEK_SDK_OVERRIDE")
```
TheButlah, line 63: This change should be reverted, as the env var that is used to override the flake input, and the env var that locates the actual sdk, are semantically different.

You can see for example how we actually use the flake input to then compute the value of the `SEEK_SDK_PATH` in the `flake.nix` [here](https://github.com/worldcoin/orb-software/blob/270fd92676799fce3af65dc88a56e7a76ecef992/flake.nix#L121) and [here](https://github.com/worldcoin/orb-software/blob/270fd92676799fce3af65dc88a56e7a76ecef992/flake.nix#L38).

Lets make this clear by naming the env var in the .envrc and CONTRIBUTING.md `FLAKE_OVERRIDE_SEEK_SDK` so its clear that it is only related to the nix flake, rather than the actual rust code.

TheButlah (changes requested):

seek-camera/sys/build.rs
AlexKaravaev: Yep, I was too fast in implementing without properly looking at full env var resolve pipeline. Now it should be consistent.

TheButlah (approved):

AlexKaravaev:
obsolete

PR #120, opened 2024-06-04
orb-slot-ctrl: Prettify handling of arguments from user

AlexKaravaev:
Noticed that parsing of status command is a bit clunky, so improved it a bit. I wanted to also add tests for this in this module, but writing tests for cli will require some more refactoring for this file, so I didn't bothered

### DIFF
--- a/orb-slot-ctrl/src/lib.rs
+++ b/orb-slot-ctrl/src/lib.rs
@@ -6,6 +6,7 @@
 use std::{
     fmt, io,
     path::{Path, PathBuf},
+    str::FromStr,
 };
 
 mod efivar;
@@ -46,8 +47,15 @@ pub enum Error {
     RemoveEfiVar { path: PathBuf, source: io::Error },
     #[error("failed reading efivar, invalid data length")]
     InvalidEfiVarLen,
+    #[error("invalid slot provided {slot}. Use one of the available slot aliases: \n{help_message}")]
+    InvalidSlotProvided { slot: String, help_message: String },
     #[error("invalid slot configuration")]
     InvalidSlotData,
+    #[error("invalid status provided {status}. Use one of the available status variant aliases: \n{help_message}")]
+    InvalidRootFsStatusProvided {
+        status: String,
+        help_message: String,
+    },
     #[error("invalid rootfs status")]
     InvalidRootFsStatusData,
     #[error("invalid retry counter({counter}), exceeding the maximum ({max})")]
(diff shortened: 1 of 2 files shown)


orb-slot-ctrl/src/lib.rs
```diff
@@ -174,6 +174,20 @@ impl TryFrom<u8> for RootFsStatus {
...
+impl TryFrom<&str> for RootFsStatus {
+    type Error = &'static str;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        match value.to_lowercase().as_str() {
+            "normal" | "0" => Ok(RootFsStatus::Normal),
+            "updateinprocess" | "updinprocess" | "1" => Ok(RootFsStatus::UpdateInProcess),
+            "updatedone" | "upddone" | "2" => Ok(RootFsStatus::UpdateDone),
+            "unbootable" | "3" => Ok(RootFsStatus::Unbootable),
+            _ => Err("Invalid status provided"),
```
TheButlah, line 186: Either use a `()` as the Err variant, or create a struct and implement thiserr on it. IMO a str should not be used.

TheButlah (changes requested):

TheButlah:
Thanks for the change! I left a minor nit, otherwise looks good.

orb-slot-ctrl/src/lib.rs
```diff
@@ -174,6 +174,20 @@ impl TryFrom<u8> for RootFsStatus {
     }
 }
 
+impl TryFrom<&str> for RootFsStatus {
```
TheButlah, line 177: The standard trait to use here is `FromStr` rather than `TryFrom`. 
You may then also be able to use this to provide a custom parser for the argument via clap.

TheButlah (changes requested):
Thanks for the contribution! A small change and should be good to go.

orb-slot-ctrl/src/lib.rs
```diff
@@ -110,6 +118,49 @@ pub enum Slot {
...
+    fn variants() -> Vec<(Slot, &'static str, Vec<&'static str>)> {
+        vec![
+            (Self::A, "A", vec!["a", "0"]),
+            (Self::B, "B", vec!["b", "1"]),
+        ]
+    }
+
+    /// Retrieves a help message listing each slot variant along with its corresponding aliases.
+    #[must_use]
+    pub fn help_message() -> String {
```
AlexKaravaev, line 131: I can move it to shared trait, but this way code grew so I wasn't sure we want things to be perfect here

orb-slot-ctrl/src/main.rs
```diff
@@ -82,120 +95,68 @@ fn main() -> eyre::Result<()> {
...
-                        println!(
-                            "{:?}",
-                            orb_slot_ctrl::get_rootfs_status(
-                                orb_slot_ctrl::get_inactive_slot()?
-                            )?
-                        );
-                    } else {
-                        println!("{:?}", orb_slot_ctrl::get_current_rootfs_status()?);
-                    }
+        Commands::Status { inactive, subcmd } => match subcmd {
```
AlexKaravaev, line 102: cargo fmt formatted this for me. are there any other configs apart from https://github.com/worldcoin/orb-software/blob/main/rustfmt.toml ?

AlexKaravaev:
> Thanks for the contribution! I also agree that using a dedicated type makes more sense. Two small requested changes and should be good to go.

I've also added clap parsing for Slot

orb-slot-ctrl/src/main.rs
TheButlah: Nope, that file is the canonical source of truth. Might be that a rust update improved the formatter? Or do you have some sort of formatting override set up by accident? CI will be the final arbiter here.

AlexKaravaev: Let me know if there are more changes needed or otherwise i will squash commits

TheButlah: No further changes needed.

TheButlah:
@AlexKaravaev can you close or bring up to date?

PR #121, opened 2024-06-05
chore: rename mcu-util->orb-mcu-util

alekseifedotov:
In LL the tool was renamed

### DIFF
--- a/orb-attest/src/remote_api.rs
+++ b/orb-attest/src/remote_api.rs
@@ -120,7 +120,7 @@ fn format_secret(val: &str) -> String {
 }
 
 fn powercycle_security_mcu() -> std::io::Result<()> {
-    Command::new("mcu-util")
+    Command::new("orb-mcu-util")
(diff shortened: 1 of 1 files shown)


orb-attest/src/remote_api.rs
```diff
@@ -120,7 +120,7 @@ fn format_secret(val: &str) -> String {
 }
 
 fn powercycle_security_mcu() -> std::io::Result<()> {
-    Command::new("mcu-util")
+    Command::new("orb-mcu-util")
```
karelispanagiotis, line 123: should we consider using full bin paths here?
cc @eth44

eth44: Great catch @karelispanagiotis, yes we want to use full paths everywhere

fouge: addressed in https://github.com/worldcoin/orb-software/pull/123

PR #122, opened 2024-06-05
mcu-interface: add example

TheButlah:
This adds an example of how to use `orb-mcu-interface`.

It also serves as a way to demonstrate that the current implementation is bugged - terminating `main` does not terminate the can_rx blocking task.

### DIFF
--- a/mcu-interface/examples/log_messages.rs
+++ b/mcu-interface/examples/log_messages.rs
@@ -1,0 +1,38 @@
+#![forbid(unsafe_code)]
+use color_eyre::{eyre::WrapErr as _, Result};
+use futures::FutureExt;
+use orb_mcu_interface::{can::canfd::CanRawMessaging, Device};
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{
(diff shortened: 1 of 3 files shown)


mcu-interface/examples/log_messages.rs
```diff
@@ -0,0 +1,38 @@
...
+        while let Ok(msg) = msg_rx.recv() {
+            println!("{msg:?}");
+        }
+    })
+    .map(|err| err.wrap_err("message task terminated unexpectedly"));
+
+    tokio::select! {
+        result = msg_task => result,
+        result = tokio::signal::ctrl_c() => { println!("ctrl-c detected"); result.wrap_err("failed to listen for ctrl-c")}
+
```
fouge, line 36: what happens if we don't have this? 
is ctrl_c still caught by the tokio handler?

fouge (approved):

mcu-interface/examples/log_messages.rs
TheButlah: if we don't have this, the rust program behaves just like any other program. The regular signal handler is invoked which terminates the process.

PR #124, opened 2024-06-06
mcu-interface: fix blocking, add kill for can_rx task, fix ack

TheButlah:
Changes in this PR:
* Turns the std::sync::mpsc channels into tokio channels. This allows us to receive messages asynchronously and reduces the number of spots we need to call blocking code from async.
* Refactors how the ack ids are created. I suspect previously there were some edge cases that could cause bugs, this new way is simpler and probably more correct.
* Adds a oneshot "kill" channel to inform `can_rx` that it should terminate.
* Places a few remaining spots that were blocking in async code behind a `tokio::task::spawn_blocking`.

### DIFF
--- a/mcu-interface/examples/log_messages.rs
+++ b/mcu-interface/examples/log_messages.rs
@@ -18,7 +18,7 @@ async fn main() -> Result<()> {
         )
         .init();
 
-    let (msg_tx, mut msg_rx) = tokio::sync::mpsc::channel(10);
+    let (msg_tx, mut msg_rx) = tokio::sync::mpsc::unbounded_channel();
     let _iface = CanRawMessaging::new(String::from("can0"), Device::Security, msg_tx)
         .wrap_err("failed to create messaging interface")?;
 
--- a/mcu-interface/examples/log_messages.rs
+++ b/mcu-interface/examples/log_messages.rs
@@ -1,6 +1,5 @@
 #![forbid(unsafe_code)]
 use color_eyre::{eyre::WrapErr as _, Result};
-use futures::FutureExt;
 use orb_mcu_interface::{can::canfd::CanRawMessaging, Device};
 use tracing::level_filters::LevelFilter;
 use tracing_subscriber::{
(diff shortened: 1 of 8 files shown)


mcu-interface/src/can/canfd.rs
```diff
@@ -96,14 +115,22 @@ impl CanRawMessaging {
...
 ) -> Result<()> {
     loop {
         let mut frame: Frame<CANFD_DATA_LEN> = Frame::empty();
         loop {
+            // terminate task on kill signal
+            use tokio::sync::oneshot::error::TryRecvError;
+            match kill_rx.try_recv() {
+                Ok(()) | Err(TryRecvError::Closed) => return Ok(()),
+                Err(oneshot::error::TryRecvError::Empty) => (),
+            }
```
fouge, line 130: makes sense

mcu-interface/src/can/mod.rs
```diff
@@ -1,2 +1,6 @@
+use std::time::Duration;
+
 pub mod canfd;
 pub mod isotp;
+
+const RX_TIMEOUT: Duration = Duration::from_millis(1500);
```
fouge, line 6: what do you think about renaming to `ACK_RX_TIMEOUT`?

mcu-interface/examples/log_messages.rs
```diff
@@ -19,19 +18,18 @@ async fn main() -> Result<()> {
         )
         .init();
 
-    let (msg_tx, msg_rx) = std::sync::mpsc::channel();
+    let (msg_tx, mut msg_rx) = tokio::sync::mpsc::channel(10);
```
fouge, line 21: is there a risk to use an `UnboundedChannel` here as well? 🤔 
messages aren't big for the Jetson.

mcu-interface/src/can/mod.rs
TheButlah: Agreed

mcu-interface/examples/log_messages.rs
TheButlah: I usually try to avoid unbounded channels whenever possible since usually you don't want memory to grow arbitrarily. I would have preferred to make the other one bounded too, but was worried about introducing some edge case that I didn't think about enough.

If we used bounded channels in both places, can you talk me through how we handle the case where the channel is full? Should we wait till it empties or warn and drop the message?

TheButlah: I could also just change to an unbounded one for now and worry about making it bounded in a subsequent PR

TheButlah:
Needs testing, will do this on Monday

mcu-interface/examples/log_messages.rs
TheButlah: Switched to unbounded.

TheButlah:
Tested orb-mcu-util and the log-messages example. both seem to work fine.

fouge (approved):

PR #125, opened 2024-06-07
mcu-util: explicitly join on task handles

TheButlah:
This PR adds explicit task handles for `can_rx`, and also propagates them up through all of the callers. It also does a bit of refactoring to mcu-util's board info logic to make it slightly more correct and easier to read.

The reason this change is needed is so that panics or other errors inside the can_rx task can be observed by the rest of the program. Otherwise, the task could silently die but the process itself and the other threads would continue happily along.

### DIFF
--- a/mcu-interface/examples/log_messages.rs
+++ b/mcu-interface/examples/log_messages.rs
@@ -19,8 +19,9 @@ async fn main() -> Result<()> {
         .init();
 
     let (msg_tx, mut msg_rx) = tokio::sync::mpsc::unbounded_channel();
-    let _iface = CanRawMessaging::new(String::from("can0"), Device::Security, msg_tx)
-        .wrap_err("failed to create messaging interface")?;
+    let (iface, task_handle) =
+        CanRawMessaging::new(String::from("can0"), Device::Security, msg_tx)
+            .wrap_err("failed to create messaging interface")?;
 
     let recv_fut = async {
         while let Some(msg) = msg_rx.recv().await {
@@ -29,8 +30,13 @@ async fn main() -> Result<()> {
     };
 
     tokio::select! {
-        () = recv_fut => Ok(()),
-        result = tokio::signal::ctrl_c() => { println!("ctrl-c detected"); result.wrap_err("failed to listen for ctrl-c")}
-
+        () = recv_fut => (),
+        result = tokio::signal::ctrl_c() => {
+            println!("ctrl-c detected");
+            result.wrap_err("failed to listen for ctrl-c")?;
+        }
     }
+
+    drop(iface);
+    task_handle.await.wrap_err("can task terminated uncleanly")
 }
--- a/mcu-interface/src/can/canfd.rs
+++ b/mcu-interface/src/can/canfd.rs
@@ -6,6 +6,7 @@ use color_eyre::eyre::{eyre, Context, Result};
 use futures::FutureExt as _;
 use orb_messages::CommonAckError;
 use prost::Message;
+use std::panic::AssertUnwindSafe;
 use std::sync::atomic::{AtomicU16, Ordering};
 use std::sync::Arc;
 use tokio::sync::{mpsc, oneshot};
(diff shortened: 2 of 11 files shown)


mcu-interface/src/can/mod.rs
fouge, line 1: a bit complex for me but if it does the job nicely.. 😁

mcu-util/src/orb/main_board.rs
```diff
@@ -364,46 +371,64 @@ impl MainBoardInfo {
...
+            Duration::from_secs(2),
+            self.listen_for_board_info(main_board),
+        )
+        .await
+        {
+            Err(tokio::time::error::Elapsed { .. }) => {
+                warn!("Timeout waiting on main board info");
+            }
+            Ok(()) => {
+                info!("got security board info");
```
fouge, line 385:
```suggestion
                debug!("got main board info");
```

mcu-util/src/orb/security_board.rs
```diff
@@ -376,34 +383,52 @@ impl SecurityBoardInfo {
...
+            Duration::from_secs(2),
+            self.listen_for_board_info(sec_board),
+        )
+        .await
+        {
+            Err(tokio::time::error::Elapsed { .. }) => {
+                warn!("Timeout waiting on security board info");
+            }
+            Ok(()) => {
+                info!("got security board info");
```
fouge, line 397:
```suggestion
                debug!("got security board info");
```

mcu-util/src/main.rs
```diff
@@ -170,16 +169,16 @@ async fn execute(args: Args) -> Result<()> {
...
         SubCommand::SecureElement(opts) => match opts {
             SecureElement::PowerCycle => {
                 orb.borrow_mut_sec_board()
                     .power_cycle_secure_element()
-                    .await
+                    .await?
             }
         },
     }
+    orb_tasks.join().await
```
fouge, line 181: this doesn't seem to return.
we would need to explicitly kill the can_rx task I think 🤔 

try with `orb-mcu-util info`, it never returns

fouge (changes requested):

mcu-interface/src/can/mod.rs
TheButlah: I agree it was complex 😅
But since we use this task handle in a lot of places, I decided it was worth it to newtype the omeshot receiver

mcu-interface/examples/log_messages.rs
```diff
@@ -29,8 +30,12 @@ async fn main() -> Result<()> {
...
-        result = tokio::signal::ctrl_c() => { println!("ctrl-c detected"); result.wrap_err("failed to listen for ctrl-c")}
-
+        () = recv_fut => (),
+        result = tokio::signal::ctrl_c() => {
+            println!("ctrl-c detected");
+            result.wrap_err("failed to listen for ctrl-c")?;
+        }
     }
+
+    task_handle.await.wrap_err("can task terminated uncleanly")
```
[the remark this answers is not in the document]
TheButlah: `shouldn't we send to the one_shot channel kill in case recv_fut returns?`
You are right. Dropping the interface will do this. I've updated the example to fix it 👍

mcu-util/src/main.rs
TheButlah: fixed

mcu-interface/src/can/canfd.rs
```diff
@@ -31,12 +32,15 @@ pub struct CanRawMessaging {
 
 impl CanRawMessaging {
     /// CanRawMessaging opens a CAN stream filtering messages addressed only to the Jetson
-    /// and start listening for incoming messages in a new blocking thread
+    /// and start listening for incoming messages in a new blocking thread.
```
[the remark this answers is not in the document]
TheButlah: mmm, in this case it actually is a thread. The messages run in a std::thread::spawn.

The thread also blocks, since .recv blocks.

fouge:
needs to be rebased apparently :)

TheButlah:
The reason why the previous commits caused things to hang is because the isotp tasks were indefinitely blocking on their receive call, and never checking the kill sender.

To work around this yesterday, I pushed some commits that:
1. use `tokio::try_join!()` to properly combine the different tasks instead of sequentially joining
2. Explicitly drop the can interfaces, to signal to the tasks that they should die
3. add a timeout to `orb_tasks.join()` to deal with the case where there were no errors while joining, but the tasks refuse to die due to the blocking receive.

fouge (approved):

PR #131, opened 2024-06-14
binaries: improve version reporting

TheButlah:
I noticed that not all binaries were including the git sha in the build information. I've added the necessary functionality to the orb_build_info crate and updated all the binaries to use it.

I also bumped the version number for orb-mcu-util.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2648,15 +2648,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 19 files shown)


TheButlah:
Rebased on #133

fouge:
there was something wrong in the rebase as there are conflicts now 🤔

fouge (approved):

PR #132, opened 2024-06-14
rust: bump to 1.79.0

TheButlah:


TheButlah:
Forced a merge because no one is online and this is a common low risk task.

PR #133, opened 2024-06-14
orb-const-concat: initial commit

TheButlah:
I didn't like that the const_format crate is so incredibly complex. There is a lot of unsafe and also a lot of cursed code in it.
I would much rather use a solution that covers the 80% use case of concatenating strings with 0 unsafe, minimal LOC, and less SBOM risk.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -967,26 +967,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "const_format"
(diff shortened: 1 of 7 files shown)


TheButlah:
Rebased on #132

fouge (approved):
I didn't expect we could to that at compile time, very cool

PR #134, opened 2024-06-14
ci: fix pull request job

TheButlah:


TheButlah:
Force merging to avoid long PR chains and because we already did this in orb-internal

PR #136, opened 2024-06-14
orb-mcu-util: don't fail on building boards object

fouge:
heartbeat is sent, when initializing orb-mcu-util, to check that microcontrollers are alive
but if it's not, don't fail because the orb-mcu-util command (ie reboot, update..) might be targeted to the other, responsive, MCU

- [x] make error cleaner
```
(venv) (os-9828b07) worldcoin@id-eff7c442:/mnt/scratch$ ./orb-mcu-util info
2024-06-14T08:04:41.249445Z ERROR orb_mcu_util::orb::security_board: Failed to send heartbeat to security mcu: Error {
    msg: "ack not received (raw)",
    source: Elapsed(
        (),
    ),
}
```

### DIFF
--- a/mcu-util/src/orb/main_board.rs
+++ b/mcu-util/src/orb/main_board.rs
@@ -66,7 +66,7 @@ impl MainBoardBuilder {
         // Send a heartbeat to the main mcu to ensure it is alive
         // & "subscribe" to the main mcu messages: messages to the Jetson
         // are going to be sent after the heartbeat
-        canfd_iface
+        let _ = canfd_iface
             .send(McuPayload::ToMain(
                 main_messaging::jetson_to_mcu::Payload::Heartbeat(
                     main_messaging::Heartbeat {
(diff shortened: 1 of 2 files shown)


TheButlah:
What does " the command might be targeted to the other MCU" mean?

fouge:
@TheButlah for example I want to reboot the security MCU, but the main MCU fails to respond: it doesn't make sense to fail

mcu-util/src/orb/main_board.rs
```diff
@@ -66,15 +66,26 @@ impl MainBoardBuilder {
...
+                if let CommonAckError::Success = c {
+                    Ok(())
+                } else {
+                    Err(eyre!("ack error: {c}"))
+                }
+            })
+            .unwrap_or_else(|e| {
+                error!("Failed to send heartbeat to main mcu: {:#?}", e);
+                Err(eyre!("Failed to send heartbeat to main mcu"))
+            });
```
TheButlah, line 88: You're producing an error just to immediately discard it.

I would rework this code to be something like:

```rust
        let ack_result = canfd_iface
            .send(McuPayload::ToMain(
                main_messaging::jetson_to_mcu::Payload::Heartbeat(
                    main_messaging::Heartbeat {
                        timeout_seconds: 0_u32,
                    },
                ),
            ))
            .await
            .map(|c| {
                if let CommonAckError::Success = c {
                    Ok(())
                } else {
                    Err(eyre!("ack error: {c}"))
                }
            });
      if let Err(e) = ack_result
           error!("Failed to send heartbeat to main mcu: {:#?}", e);
      });
```

TheButlah (changes requested):

mcu-util/src/orb/main_board.rs
TheButlah: I pushed a commit for this just now.

TheButlah (approved):

TheButlah:
I'm not entirely sure this is the right approach, but I am going to merge for now.

PR #137, opened 2024-06-17
ci: fix tmp deletion

TheButlah:


TheButlah:
didnt help, I guess the action is just broken or I have a skill issue.

PR #143, opened 2024-07-02
orb-ui: failures reasons all into SignupFail event

fouge:
instead of having some reasons outside of this event

this is breaking the API but I think it's fine because versions are hardcoded
we need to make sure that the new versions of orb-core and orb-ui are changed
in the same orb-os commit

### DIFF
--- a/orb-ui/src/engine/pearl.rs
+++ b/orb-ui/src/engine/pearl.rs
@@ -603,6 +603,18 @@ impl EventHandler for Runner<PEARL_RING_LED_COUNT, PEARL_CENTER_LED_COUNT> {
                             sound::Voice::VerificationNotSuccessfulPleaseTryAgain,
                         ))?;
                     }
+                    SignupFailReason::SoftwareVersionDeprecated => {
+                        self.operator_blink.trigger(
+                            Argb::PEARL_OPERATOR_VERSIONS_DEPRECATED,
+                            vec![0.4, 0.4, 0.4, 0.4, 0.4, 0.4],
+                        );
+                    }
+                    SignupFailReason::SoftwareVersionBlocked => {
+                        self.operator_blink.trigger(
+                            Argb::PEARL_OPERATOR_VERSIONS_OUTDATED,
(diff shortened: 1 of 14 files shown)


orb-ui/src/engine/pearl.rs
```diff
@@ -603,6 +603,18 @@ impl EventHandler for Runner<PEARL_RING_LED_COUNT, PEARL_CENTER_LED_COUNT> {
                             sound::Voice::VerificationNotSuccessfulPleaseTryAgain,
                         ))?;
                     }
+                    SignupFailReason::SoftwareVersionDeprecated => {
+                        self.operator_blink.trigger(
+                            Argb::PEARL_OPERATOR_VERSIONS_DEPRECATED,
+                            vec![0.4, 0.4, 0.4, 0.4, 0.4, 0.4],
```
TheButlah, line 609:
```suggestion
                            vec![0.4; 6],
```

orb-ui/src/engine/pearl.rs
```diff
@@ -603,6 +603,18 @@ impl EventHandler for Runner<PEARL_RING_LED_COUNT, PEARL_CENTER_LED_COUNT> {
...
+                        );
+                    }
+                    SignupFailReason::SoftwareVersionBlocked => {
+                        self.operator_blink.trigger(
+                            Argb::PEARL_OPERATOR_VERSIONS_OUTDATED,
+                            vec![0.4, 0.4, 0.4, 0.4, 0.4, 0.4],
```
TheButlah, line 615:
```suggestion
                            vec![0.4; 6],
```

TheButlah (changes requested):

orb-ui/src/engine/pearl.rs
fouge: it's more rusty but less easy visually to count the number of blinks 😁

TheButlah (dismissed):

TheButlah (approved):

PR #144, opened 2024-07-02
mcu-interface: isotp: set block size to 8

fouge:
when the MCU tries to send payloads that are
larger than 7 bytes, ISO-TP chunks the data.
The block size is set by the receiver and
indicates to the sender when a flow control
packed must be awaited.
8 CF (consecutive frames) are now sent
by the sender before one FC (flow control)
frame is sent back.


TheButlah:
still relevant or should we close?

fouge:
adding to backlog

PR #146, opened 2024-07-02
orb-mcu-util: security mcu to apply update directly

fouge:
and better wording

### DIFF
--- a/mcu-util/src/orb/main_board.rs
+++ b/mcu-util/src/orb/main_board.rs
@@ -203,18 +203,18 @@ impl Board for MainBoard {
         if let Ok(ack) = self.isotp_iface.send(payload).await {
             if !matches!(ack, CommonAckError::Success) {
                 return Err(eyre!(
-                    "Unable to validate image: ack error: {}",
+                    "Unable to check image integrity: ack error: {}",
                     ack as i32
                 ));
             }
             info!("✅ Image integrity confirmed, activating image");
         } else {
-            return Err(eyre!("Firmware image check failed"));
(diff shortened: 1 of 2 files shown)


mcu-util/src/orb/main_board.rs
```diff
@@ -210,18 +203,18 @@ impl Board for MainBoard {
...
             info!("✅ Image integrity confirmed, activating image");
         } else {
-            return Err(eyre!("Firmware image check failed"));
+            return Err(eyre!("Firmware image integrity check failed"));
         }
 
         self.switch_images().await?;
 
-        info!("👉 Shut the Orb down to install the new image");
+        info!("👉 Shut the Orb down to install the new image (`sudo shutdown now`), the Orb is going to reboot itself once installation is complete");
```
TheButlah, line 217: Just as a heads up - the supervisor daemon provides a way to shutdown/reboot without root. Consult [here](https://github.com/worldcoin/orb-internal/blob/710f1f6f546c972cd179bbadba41c0c8ac9f26eb/orb-mcu-telemetry/src/main.rs#L177-L181) for prior art.

TheButlah (approved):

mcu-util/src/orb/main_board.rs
fouge: you mean with a dbus call that's performed from orb-mcu-util? 

I could add that command maybe 🤔

TheButlah: Yep

PR #152, opened 2024-07-10
orb-ui: update diamond ui

fouge:
update diamond UI with latest specs.

- latest colors
- Alert now with smooth transitions instead of binary states. Allows the new "capture succeeded" animation.
- refacto: all animations (center and ring) now into `animations`. This removes duplicate
- fix all tests
- added test for Alert


orb-ui/src/engine/center/alert.rs
```diff
@@ -24,10 +32,16 @@ pub struct Alert<const N: usize> {
 impl<const N: usize> Alert<N> {
     /// Creates a new [`Alert`].
     #[must_use]
-    pub fn new(color: Argb, pattern: Vec<f64>, start_on: bool) -> Self {
+    pub fn new(
+        color: Argb,
+        pattern: Vec<f64>,
```
TheButlah, line 37: Newtype this vec for added type safety.

orb-ui/src/engine/center/alert.rs
```diff
@@ -24,10 +32,16 @@ pub struct Alert<const N: usize> {
...
+        smooth_transitions: Option<Vec<f64>>,
+        start_on: bool,
```
TheButlah, line 39: Can you rename to "leds_active_at_start" or something like that so that its more clear what this is?

orb-ui/src/engine/center/alert.rs
```diff
@@ -53,25 +67,49 @@ impl<const N: usize> Animation for Alert<N> {
...
+                let period = self.pattern[i + 1] - smooth;
+                let t = self.phase - time_acc - smooth;
+                let intensity = if self.current_solid_color == Argb::OFF {
+                    // starts at intensity 0
+                    1.0 - (t * (PI / 2.0) / period).cos()
+                } else {
+                    // starts at intensity 1
+                    (t * (PI / 2.0) / period).cos()
+                };
+                color = self.target_color * intensity;
```
TheButlah, line 106: @fouge can you add a unit test for some of this math? To just check a few values that you expect. This would give me more confidence in the math. Also, giving this a function name with a couple tests may make it easier for me to understand what its doing.

orb-ui/src/engine/diamond.rs
```diff
@@ -288,15 +289,17 @@ impl EventHandler for Runner<DIAMOND_RING_LED_COUNT, DIAMOND_CENTER_LED_COUNT> {
                     ring::alert::Alert::<DIAMOND_RING_LED_COUNT>::new(
                         Argb::DIAMOND_USER_QR_SCAN,
                         vec![0.0, 0.3, 0.3],
-                        false,
+                        None,
+                        true,
```
TheButlah, line 293: Some food for thought: It would be more apparent what `true` means at the callsite, if you used an enum or an explicit type.

orb-ui/src/engine/ring/alert.rs
```diff
@@ -52,25 +66,49 @@ impl<const N: usize> Animation for Alert<N> {
...
+                let period = self.pattern[i + 1] - smooth;
+                let t = self.phase - time_acc - smooth;
+                let intensity = if self.current_solid_color == Argb::OFF {
+                    // starts at intensity 0
+                    1.0 - (t / period * (PI / 2.0)).cos()
+                } else {
+                    // starts at intensity 1
+                    (t * (PI / 2.0) / period).cos()
+                };
+                color = self.target_color * intensity;
```
TheButlah, line 105: same here, can you add a unit test?

TheButlah (changes requested):

orb-ui/src/engine/ring/alert.rs
fouge: I refactored so that we only have one `alert.rs` 
 the code is the same
I also added a test

TheButlah (approved):

orb-ui/src/engine/animations/alert.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+impl<const N: usize> Animation for Alert<N> {
+    type Frame = [Argb; N];
+
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn as_any_mut(&mut self) -> &mut dyn Any {
```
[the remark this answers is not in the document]
fouge: have a look at the Trait and usages in the code.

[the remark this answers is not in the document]
TheButlah: I think what I mean is: It makes more sense to use `dyn Animation` (which is already what is stored in the animation stack) than `dyn Any`. If you want the ability to downcast animations, IMO it would be better to create a `downcast_ref()` and `downcast_mut()` function on the Animation trait, which in its default impl, calls std::any::Any::downcast_ref to do the downcast.

The reason this is prefereable over using dyn Any is that dyn Animation is a more narrow type, and that it makes the code a bit easier to understand.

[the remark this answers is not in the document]
fouge: tbh I didn't write that piece of code. it was taken from orb-core.
I am not really sure how to implement this, can you propose a PR if you think it deserves it?
thanks

TheButlah (approved):

PR #153, opened 2024-07-11
orb-ui: diamond self-serve with cone

fouge:
initial ftdi driver.
create devices for LCD, LED and button.
button events sent to a channel.
led strip driver.
with example

fixes O-2793


fouge:
@TheButlah I am setting this PR as ready to review even though the UI is not final, later work is gonna be done in separate PRs.

fouge:
see #170 #171 #172

PR #155, opened 2024-07-16
security: GTS_Root_R2/R3/R4 certs

oldgalileo:
Sourced from 'https://pki.goog', fingerprints checked.


vmenge (approved):

TheButlah (approved):

TheButlah:
Note that these changes will not take affect until all the binaries are updated qwq

PR #156, opened 2024-07-16
hil: add ftdi reboot support

TheButlah:
Integrates with libftd2xx (proprietary but ok for us to use) to control the GPIO pins on the ftdi adapter from orb-hil. This allows us to trigger reboots and recovery mode.

Also introduces an explicit reboot subcommand.


deny.toml
```diff
@@ -38,3 +39,16 @@ expression = "LicenseRef-ring"
 license-files = [
     { path = "LICENSE", hash = 0xbd0eed23 },
 ]
+
+[[licenses.clarify]]
+# @thebutlah reviewed the license code. It doesn't impose any copyleft
+# restrictions on us, and doesn't seem to impose any restrictions on
+# distribution. As long as we don't put any other copyleft code in
+# orb-software, we won't run afoul of any license restrictions AFAICT. We
+# should seek to reimplement pyftdi in rust using nusb if we can, though.
```
fouge, line 48: we could try that https://github.com/tanriol/libftdi1-sys 🤔

TheButlah: That library is LGPL, so we cannot use it.

hil/src/main.rs
```diff
@@ -52,45 +128,17 @@ async fn main() -> Result<()> {
...
 
     let args = Cli::parse();
-
-    let rts_path = if let Some(ref s3_url) = args.s3_url {
-        if args.rts_path.is_some() {
-            bail!("both rts_path and s3_url were specified - only provide one or the other");
+    let run_fut = async {
+        match args.commands {
+            Commands::Flash(c) => c.run().await,
+            Commands::Reboot(c) => c.run().await,
```
fouge, line 134: I like this. I should to the same for orb-mcu-util 🤔

fouge (changes requested):
I don't think i went through all of it.

hil/Cargo.toml
```diff
@@ -12,15 +12,18 @@ rust-version.workspace = true
 
 [dependencies]
 camino = "1.1.6"
-clap = { version = "4", features = ["derive"] }
+clap = { workspace = true, features = ["derive"] }
 cmd_lib = "1.9.3"
-color-eyre = "0.6"
+color-eyre.workspace = true
+ftdi-embedded-hal = { version = "0.22.0", features = ["libftd2xx-static"] }
```
fouge, line 18: should we set this dependency at the workspace level because the cone is also using it. 
otherwise i'm gonna do it in my PR

hil/src/ftdi.rs
```diff
@@ -0,0 +1,287 @@
...
+            desired_state: current_pin_state,
+            serial,
+            is_destroyed: false,
+        })
+    }
+}
+
+/// An FTDI device configured in Async GPIO bitbang mode.
+///
+/// You can use this for controlling the pings of the FTDI adapter like a GPIO device.
```
fouge, line 131:
```suggestion
/// You can use this for controlling the pins of the FTDI adapter like a GPIO device.
```

fouge (approved):

fouge (approved):

oldgalileo (approved):
🏃 LGTM

PR #165, opened 2024-07-22
security-utils: min tls ver to 1.2

vmenge:


karelispanagiotis (approved):
LGTM! tested that it fixes the issue of failing to upload to s3 by doing some signups.

PR #166, opened 2024-07-26
scripts: Script to fuse random orb-id for dev orb

karelispanagiotis:
Introducing `dev-orb-setup` scripts to make our lives easier when setting up a fresh dev orb.

This PR add a new script that fuses a random orb-id on the orb's jetson `reserved_odm0` fuses.

### DIFF
--- a/scripts/dev-orb-setup/fuse-orb-id.sh
+++ b/scripts/dev-orb-setup/fuse-orb-id.sh
@@ -10,15 +10,18 @@ BASE_DIR=$(cd -- "$(dirname -- "${SCRIPT_DIR}")" &> /dev/null && pwd)
 
 usage() {
     echo "Description: Script to fuse a random ORB-ID"
-    echo "Usage: $0 <orb-ip-address>"
+    echo "Usage: $0 <orb-ip-address> [-s <ssh-socket-name>]"
     echo ""
     echo "Arguments:"
-    echo "  <orb-ip-address>    The IP address of the orb."
+    echo "  <orb-ip-address>              The IP address of the orb."
+    echo "  -s | --ssh-socket <name>      Optional SSH socket name."
 }
+
 
 main() {
     local arg
     local positional_args=()
+    local ssh_socket="tmp-ssh-socket"
 
     while [[ $# -gt 0 ]]; do
         arg="${1}"; shift
@@ -26,6 +29,9 @@ main() {
             -h | --help)
                 usage
                 exit 0
+                ;;
+            -s | --ssh-socket)
+                ssh_socket="${1}"; shift
                 ;;
             -*)
                 echo "Invalid argument: ${arg}"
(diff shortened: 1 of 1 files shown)


scripts/dev-orb-setup/fuse-orb-id.sh
```diff
@@ -0,0 +1,77 @@
...
+        exit 1
+    fi
+
+    local orb_ip="${1}"
+    local lock_status
+    local orb_id
+
+    ssh -M -S tmp-ssh-socket -fN worldcoin@"${orb_ip}"
+
+    lock_status=$(ssh -S tmp-ssh-socket worldcoin@"${orb_ip}" 'cat /sys/devices/platform/tegra-fuse/odm_lock')
```
oldgalileo, line 58: Nit: All the `ssh` commands can fail in cases related to the number of keys configured on the host.

Because you're using the same socket, I _**think**_ it's already authed and you don't need to reauth on each subsequent command.

scripts/dev-orb-setup/fuse-orb-id.sh
```diff
@@ -0,0 +1,77 @@
...
+        echo "Error: Exactly one positional argument is required."
+        usage
+        exit 1
+    fi
+
+    local orb_ip="${1}"
+    local lock_status
+    local orb_id
+
+    ssh -M -S tmp-ssh-socket -fN worldcoin@"${orb_ip}"
```
oldgalileo, line 56: I would add an optional argument to set a different path than `tmp-ssh-socket`. This way, each script called from the master script can re-use the same SSH connection.

scripts/dev-orb-setup/fuse-orb-id.sh
```diff
@@ -0,0 +1,77 @@
...
+
+    lock_status=$(ssh -S tmp-ssh-socket worldcoin@"${orb_ip}" 'cat /sys/devices/platform/tegra-fuse/odm_lock')
+
+    if [[ "${lock_status}" == "0x00000001" ]]; then
+        echo "orb-id is already fused."
+        ssh -S tmp-ssh-socket -O exit worldcoin@"${orb_ip}"
+        exit 0
+    fi
+
+    orb_id=$(openssl rand -hex 4)
```
oldgalileo, line 66:
```suggestion
    orb_id="$(openssl rand -hex 4)"
```

scripts/dev-orb-setup/fuse-orb-id.sh
oldgalileo, line 58:
```suggestion
    lock_status="$(ssh -S tmp-ssh-socket worldcoin@"${orb_ip}" 'cat /sys/devices/platform/tegra-fuse/odm_lock')"
```

scripts/dev-orb-setup/fuse-orb-id.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
```
oldgalileo, line 1:
```suggestion
#!/usr/bin/env bash
```

oldgalileo (changes requested):
Proud. Good bash script. Readable 👍 

Some nits, some small things.

scripts/dev-orb-setup/fuse-orb-id.sh
karelispanagiotis: done

scripts/dev-orb-setup/fuse-orb-id.sh
karelispanagiotis: you mean I can omit the `worldcoin@"${orb_ip}"` in all of the `ssh` commands?

PR #170, opened 2024-07-31
orb-rgb: new crate to manage LED colors and transformations

fouge:


TheButlah (approved):
Approved. Just curious, what motivated the new crate?

PR #171, opened 2024-07-31
orb-ui: cone library to control button, leds and display

fouge:
drivers for cone components that are connected to an FTDI chip that acts as a USB-accessible bridge between the Jetson and the components:
- button: polled every 50ms
- led strip: APA102 LEDs controllable via a spi bus 
- display: lcd with Gc9a01 driver, that is used to display images, qr codes, or full-screen colors.


orb-ui/cone/src/led.rs
```diff
@@ -0,0 +1,140 @@
...
+use color_eyre::eyre::Context;
+use ftdi_embedded_hal::eh1::spi::SpiBus;
+use ftdi_embedded_hal::libftd2xx::{Ft4232h, Ftdi, FtdiCommon};
+use orb_rgb::Argb;
+use std::sync::{Arc, Mutex};
+use tokio::sync::{mpsc, watch};
+use tokio::task;
+
+pub struct Led {
+    spi: ftdi_embedded_hal::Spi<Ft4232h>,
```
TheButlah, line 11: `spi` should be owned by the task, not by the proxy struct. The point of the task is to provide a single spot that has exclusive access to the hardware, and the proxy object communicates with it and manages the task's lifetime (and optionally provides RPC or a way to get a channel to communicate with it).

Most probably, if you move `spi` into the task, you will no longer need the Arc<Mutex>.

orb-ui/cone/src/led.rs
```diff
@@ -0,0 +1,140 @@
...
+        tracing::debug!("LED strip initialized");
+
+        Ok(led)
+    }
+
+    pub(crate) fn clone_tx(&self) -> mpsc::UnboundedSender<[Argb; CONE_LED_COUNT]> {
+        self.led_strip_tx.clone()
+    }
+
+    fn spi_rgb_led_update(&mut self, buffer: &[u8]) -> eyre::Result<()> {
```
TheButlah, line 78: I would probably make this a helper function that is not associated to any struct. Definitely doesn't go on `Led` though, because this is executed from the task and `Led` should not be accessible from the task (avoids an Arc).

TheButlah, line 78: This should take an `impl SpiBus` as an argument. The reason you would use the impl is so that we can provide a mock io Spi device and test this code more easily.

additionally, please provide one unit test by mocking an SpiBus using a vec, and checking that the data written to the vec is as expected.

orb-ui/cone/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Create a new Cone instance.
+    pub fn new(event_queue: mpsc::UnboundedSender<ConeEvents>) -> eyre::Result<Self> {
+        let connection_status = if ftdi_embedded_hal::libftd2xx::list_devices()?.len()
+            != CONE_FTDI_DEVICE_COUNT
+        {
+            return Err(eyre::eyre!(
+                "FTDI device count mismatch: cone not connected?"
+            ));
+        } else {
+            let mut device: Ft4232h = Ftdi::with_index(6)?
```
TheButlah, line 48:
```suggestion
            // TODO: Figure out a better way to select a specific FTDI device.
            let mut device: Ft4232h = Ftdi::with_index(6)?
```

orb-ui/cone/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+    _button: Button,
+}
+
+pub enum ConeEvents {
+    ButtonPressed(bool),
+}
+
+impl Cone {
+    /// Create a new Cone instance.
+    pub fn new(event_queue: mpsc::UnboundedSender<ConeEvents>) -> eyre::Result<Self> {
```
TheButlah, line 40: I suspect this may want to be a broadcast channel instead of a mpsc. Do we anticipate multiple people reading events?

orb-ui/cone/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+            connection_status,
+            lcd,
+            led_strip,
+            _button,
+        };
+
+        Ok(cone)
+    }
+
+    pub fn is_connected(&self) -> bool {
```
TheButlah, line 69: I'm not sure that sharing an arc mutex and storing it in the cone struct like this is a good idea. Probably makes more sense to have Connected and disconnected events that are sent through the channel.

Alternatively, use an arc [atomic enum](https://docs.rs/atomic_enum/latest/atomic_enum/) or an `Arc<AtomicBool>`. You won't have to think about deadlocks or poisoning that way.

orb-ui/cone/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    /// Update the RGB LEDs by passing the values to the LED strip sender.
+    pub fn queue_rgb_leds(
+        &mut self,
+        pixels: &[Argb; CONE_LED_COUNT],
+    ) -> eyre::Result<()> {
+        self.led_strip
+            .lock()
+            .expect("cannot lock LED strip mutex")
+            .clone_tx()
```
TheButlah, line 85: If you don't use an Arc<Mutex<Led>>, the &mut self will allow you to avoid the lock and avoid the clone_tx.

orb-ui/cone/src/led.rs
```diff
@@ -0,0 +1,140 @@
...
+        if let Ok(led) = &mut led.lock() {
+            led.task_handle = Some(task);
+        }
+
+        tracing::debug!("LED strip initialized");
+
+        Ok(led)
+    }
+
+    pub(crate) fn clone_tx(&self) -> mpsc::UnboundedSender<[Argb; CONE_LED_COUNT]> {
```
TheButlah, line 74: Expose it as a reference, instead of forcing everyone to always clone.

orb-ui/cone/src/lib.rs
TheButlah, line 40: Should take all of the JoinHandles of the various tasks, and aggregate them into a struct, and return that as part of a tuple.

Also, lets call it `spawn` instead of `new` so its more clear that we are running a bunch of tasks.

```rust
#[derive(Debug)]
struct ConeJoinHandles {
  pub lcd: LcdJoinHandle,
  pub led: LedJoinHandle,
  pub button: ButtonJoinHandle,
}

impl ConeJoinHandles {
  async fn join(self) -> eyre::Result<()> {
    let _: ((),(),()) = tokio::try_join!(
      // ...
    )?;

    Ok(())
  }
}
```
See [what we did in mcu-util](https://github.com/worldcoin/orb-software/blob/9e2e3908c81de12f95b5a69b8a7fefd3baa0b209/mcu-util/src/orb/mod.rs#L264-L298) for inspiration.

orb-ui/cone/src/button.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Poll the button state.
+/// Events are sent to the event queue when the button is pressed or released
+/// The thread that polls the button is also used to check the connection status
+impl Button {
+    pub(crate) fn spawn(
+        event_queue: mpsc::UnboundedSender<ConeEvents>,
+        connection: Arc<Mutex<Status>>,
+    ) -> eyre::Result<Self> {
+        let mut device: Ft4232h = Ftdi::with_index(7)?.try_into()?;
```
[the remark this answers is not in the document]
fouge: I centralized all the indexes into `lib.rs` and added the comment

orb-ui/cone/src/button.rs
```diff
@@ -0,0 +1,86 @@
...
+    terminate: Arc<AtomicBool>,
+}
+
+/// Poll the button state.
+/// Events are sent to the event queue when the button is pressed or released
+/// The thread that polls the button is also used to check the connection status
+impl Button {
+    pub(crate) fn spawn(
+        event_queue: mpsc::UnboundedSender<ConeEvents>,
+        connection: Arc<Mutex<Status>>,
```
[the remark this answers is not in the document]
fouge: it's a leftover from me trying to make orb-ui able to use the cone if dis/connected at runtime.
I am removing it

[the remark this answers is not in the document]
fouge: but i agree I should add the event as a `ConeEvent`

orb-ui/cone/src/button.rs
```diff
@@ -0,0 +1,86 @@
...
+    ) -> eyre::Result<Self> {
+        let mut device: Ft4232h = Ftdi::with_index(7)?.try_into()?;
+        let mask: u8 = !BUTTON_GPIO_MASK; // button pin as input, all others as output
```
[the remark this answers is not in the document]
fouge: I defined a new const `BUTTON_GPIO_PINS_IO_MASK`

orb-ui/cone/src/button.rs
```diff
@@ -0,0 +1,86 @@
...
+            // keep state so that we send an event only on state change
+            let mut last_state = false;
+            loop {
+                if terminate_clone.load(Ordering::Relaxed) {
+                    return;
+                }
+                match device.bit_mode() {
+                    Ok(mode) => {
+                        // connected
+                        if let Ok(mut status) = connection.lock() {
```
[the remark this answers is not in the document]
fouge: I removed the mutex

orb-ui/cone/src/lib.rs
fouge: done

orb-ui/cone/src/lib.rs
fouge: done

orb-ui/cone/src/lcd.rs
```diff
@@ -0,0 +1,186 @@
...
+        Ok(Lcd {
+            tx,
+            shutdown_signal,
+            task_handle: Some(task_handle),
+        })
+    }
+
+    pub(crate) fn clone_tx(&self) -> mpsc::UnboundedSender<LcdCommand> {
+        self.tx.clone()
+    }
```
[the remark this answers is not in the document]
fouge: I changed a bit how things work: I added a function called `send` that uses the internal queue to pass the new command to the LCD task. let me know what you think

orb-ui/cone/src/led.rs
fouge: I created a new struct called `Apa102` and that's also because the function is quite dependent on which LED is used

orb-ui/cone/src/lcd.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+async fn handle_lcd_update(
+    rx: &mut mpsc::UnboundedReceiver<LcdCommand>,
+    mut shutdown_receiver: Receiver<()>,
+) -> eyre::Result<()> {
+    let mut delay = Delay::new();
```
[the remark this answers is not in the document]
fouge: done [35b06d2](https://github.com/worldcoin/orb-software/pull/171/commits/35b06d281e3bec9b298b388bf2ed36048fd72a66)

orb-ui/cone/src/button.rs
```diff
@@ -1,86 +1,90 @@
...
-                        if pressed != last_state {
-                            if let Err(e) =
-                                event_queue.send(ConeEvents::ButtonPressed(pressed))
-                            {
-                                tracing::error!("Error sending event: {:?} - no receiver? stopping producer", e);
-                                return;
+                select! {
+                    _ = &mut kill_rx => return Ok(()),
+                    _ = tokio::time::sleep(std::time::Duration::from_millis(BUTTON_POLL_INTERVAL_MS)) => {
+                        match device.bit_mode() {
```
TheButlah, line 61: you're blocking in async. use tokio::task::spawn_blocking instead of spawn(). The first attempt used std::thread which is why I never complained about that part :p

orb-ui/cone/src/lcd.rs
```diff
@@ -8,30 +9,32 @@ use ftdi_embedded_hal::libftd2xx::{Ft4232h, Ftdi, FtdiCommon};
...
 /// The task is spawned when the Lcd is created
 /// and stopped when the Lcd is dropped
 pub struct Lcd {
-    tx: mpsc::UnboundedSender<LcdCommand>,
-    shutdown_signal: watch::Sender<()>,
-    task_handle: Option<JoinHandle<eyre::Result<()>>>,
+    /// Used to signal that the task should be cleanly terminated.
+    pub kill_tx: oneshot::Sender<()>,
+    /// Send commands to the LCD task
+    cmd_tx: mpsc::UnboundedSender<LcdCommand>,
```
TheButlah, line 33: imo it should be bounded. The backpressure to the sender is important.

orb-ui/cone/src/lcd.rs
```diff
@@ -40,33 +43,28 @@ pub enum LcdCommand {
...
-            task_handle: Some(task_handle),
-        })
+            task::spawn(async move { do_lcd_update(&mut cmd_rx, kill_rx).await });
+
+        Ok((Lcd { cmd_tx, kill_tx }, LcdJoinHandle(task_handle)))
     }
 
-    pub(crate) fn clone_tx(&self) -> mpsc::UnboundedSender<LcdCommand> {
-        self.tx.clone()
+    pub(crate) fn send(&mut self, cmd: LcdCommand) -> eyre::Result<()> {
```
TheButlah, line 56: Just expose it as a ref. Gives a richer, more powerful api. Especially important if you switch to a bounded channel, as you'll need access to both send and try send, blocking send, etc.

Also, it should only require &self.

orb-ui/cone/src/lcd.rs
```diff
@@ -96,20 +94,18 @@ async fn handle_lcd_update(
...
 
     loop {
-        tokio::select! {
-            _ = shutdown_receiver.changed() => {
-                debug!("LCD task shutting down");
-                return Ok(())
+        select! {
+            _ = &mut kill_rx => {
+                // gracefully terminate
+                tracing::trace!("lcd task killed");
```
TheButlah: How I think about log levels:

info: informational messages that do get sent to datadog
debug: informational messages that don't get sent to datadog by default
trace: high frequency spam

IMO this falls into the "debug" category.

TheButlah (changes requested):

orb-ui/cone/src/led.rs
```diff
@@ -0,0 +1,248 @@
...
+            self.written.borrow_mut().extend_from_slice(words);
+            Ok(())
+        }
+
+        fn transfer(
+            &mut self,
+            _read: &mut [u8],
+            _write: &[u8],
+        ) -> Result<(), Self::Error> {
+            todo!()
```
TheButlah, line 184: use `unimplemetned` for things you intentionally did not implement, use `todo` for things that you intend to do in the future

orb-ui/cone/examples/cone-simulation.rs
```diff
@@ -0,0 +1,188 @@
...
+    for device in devices.iter() {
+        tracing::debug!("Device: {:?}", device);
+    }
+
+    loop {
+        let (tx, mut rx) = broadcast::channel(10);
+        if let Ok((mut cone, cone_handles)) = orb_cone::Cone::spawn(tx) {
+            // spawn a thread to receive events
+            let button_listener_task: JoinHandle<eyre::Result<()>> =
+                task::spawn(async move {
```
TheButlah, line 46: nit: make the async block into an async fn to reduce the indentation

orb-ui/cone/examples/cone-simulation.rs
```diff
@@ -0,0 +1,188 @@
...
+            let (kill_sim_tx, mut kill_sim_rx) = tokio::sync::oneshot::channel::<()>();
+            let simulation_task: JoinHandle<eyre::Result<()>> = tokio::task::spawn(
+                async move {
+                    let mut counter = 0;
+                    loop {
+                        tokio::select! {
+                            _ = &mut kill_sim_rx => {
+                                return Ok(());
+                            }
+                            _ = tokio::time::sleep(std::time::Duration::from_secs(CONE_SIMULATION_UPDATE_PERIOD_S)) => {
```
TheButlah, line 83: nit: reduce indents by making the stuff after => happen *after* the select is finished. Same concept as doing 
```rust
if is_bad {
  bail!("oh no");
}
do_good();
```

rather than
```rust
if is_bad {
  bail!("oh no");
} else {
  do_good()
}
```

orb-ui/cone/examples/cone-simulation.rs
```diff
@@ -0,0 +1,188 @@
...
+                            Err(RecvError::Lagged(skipped)) => {
+                                tracing::warn!("🚨 Skipped {} cone events", skipped);
+                            }
+                        }
+                    }
+                });
+
+            // create one shot to gracefully terminate simulation
+            let (kill_sim_tx, mut kill_sim_rx) = tokio::sync::oneshot::channel::<()>();
+            let simulation_task: JoinHandle<eyre::Result<()>> = tokio::task::spawn(
```
TheButlah, line 75: nit: simulation task could just be an async fn instead of a task. This would simplify things because you no longer need to have kill signals or think about the tasks lifecycle independently from the main function's future.

I don't have concrete guidance on when to make something a task vs just a regular future. But I *think* that here, a future makes more sense.

orb-ui/cone/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "orb-cone"
+version = "0.0.0"
+authors = ["Cyril Fougeray"]
```
TheButlah, line 4: the other parts of the code include the email, did you intend to omit it?

orb-ui/cone/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+publish = false
+
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
TheButlah, line 12:
```suggestion
```

orb-ui/cone/examples/cone-simulation.rs
```diff
@@ -0,0 +1,188 @@
...
+                                        ))?;
+                                        for pixel in pixels.iter_mut() {
+                                            *pixel = Argb::DIAMOND_USER_AMBER;
+                                        }
+                                    }
+                                    _ => {}
+                                }
+                                cone.queue_rgb_leds(&pixels)?;
+                            }
+                        } // end tokio::select!
```
TheButlah, line 147: This is often a sign that the thing inside the {} should be broken into a separate function or closure :)

orb-ui/cone/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+            lcd: lcd_handle,
+            led_strip: led_handle,
+            button: button_handle,
+        };
+
+        Ok((cone, handle))
+    }
+
+    /// Update the RGB LEDs by passing the values to the LED strip sender.
+    pub fn queue_rgb_leds(
```
TheButlah, line 121: summary of zoom meeting: I pointed out how the error type here will treat a full channel as an error, and make it hard for the caller to handle that case.

Instead, we think directly exposing the channel and the LcdCommand enum would be nicer. Then you can do this, and delete all the `queue_*` functions:
```rust
let _ = cone.lcd().tx().try_send(LcdCommand::rgb_leds(pixels)?));
```

I.e. we separate the type conversion/validation out, directly exposing the underlying communication pattern.

orb-ui/cone/src/led.rs
```diff
@@ -0,0 +1,248 @@
...
+                // todo do we want to only take the last message and ignore previous ones
+                let msg = rt.block_on(async {
+                    tokio::select! {
+                        _ = &mut kill_rx => {
+                            tracing::trace!("led task killed");
+                            None
+                        }
+                        msg = rx.recv() => msg,
+                    }
+                });
```
TheButlah, line 57: yeet

orb-ui/cone/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+impl ConeJoinHandle {
+    pub async fn join(self) -> eyre::Result<()> {
+        let (lcd_e, led_e, button_e) =
+            tokio::try_join!(self.lcd.0, self.led_strip.0, self.button.0)?;
+
+        // print any error that occurred
+        if let Err(e) = lcd_e {
+            tracing::error!("LCD task error: {:?}", e);
```
TheButlah, line 46: Why not just return the error to the caller? Like we do here:
https://github.com/worldcoin/orb-software/blob/17e515bd1c1f91c2ac4a06bd44b9d1791f601ddf/orb-backend-state/src/main.rs#L110-L119

orb-ui/cone/src/led.rs
```diff
@@ -0,0 +1,248 @@
...
+
+/// LED strip handle.
+/// To send new values to the LED strip.
+pub struct LedStrip {
+    /// Used to signal that the task should be cleanly terminated.
+    pub kill_tx: oneshot::Sender<()>,
+    tx: mpsc::Sender<[Argb; CONE_LED_COUNT]>,
+}
+
+pub struct LedJoinHandle(pub task::JoinHandle<eyre::Result<()>>);
```
TheButlah, line 20: Always #[derive(Debug)]. Should probably turn on that lint in clippy.

TheButlah (changes requested):

orb-ui/cone/src/lib.rs
fouge: I initially wanted to only print a message when a task exit with an error to debug.

I want to apply the same pattern but I can't use `map()` on `self.lcd.0` 🤔

orb-ui/cone/src/led.rs
fouge: why? 🤔 
should I do it on `LedStrip` above as well?

orb-ui/cone/src/lib.rs
TheButlah: You need the `futures::FutureExt` [extension trait](https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html#what-is-an-extension-trait).

TheButlah (approved):

TheButlah (approved):

PR #176, opened 2024-08-07
supervisor: added .service file

TheButlah:
As part of merging orb-supervisor into orb-software, we need to make it's .deb file include the service file. Previously this lived in orb-os, but to be consistent with the way we do other stuff, I believe it should be moved into orb-software.

There is an associated PR at worldcoin/orb-os#478

### DIFF
--- a/orb-supervisor/Cargo.toml
+++ b/orb-supervisor/Cargo.toml
@@ -34,3 +34,10 @@ dbus-launch = "0.2.0"
 tokio = { version = "1.25.0", features = ["sync", "test-util"] }
 
 [package.metadata.deb]
+maintainer-scripts = "debian/"
+assets = [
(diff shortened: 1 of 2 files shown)


orb-supervisor/Cargo.toml
```diff
@@ -34,3 +34,10 @@ dbus-launch = "0.2.0"
 tokio = { version = "1.25.0", features = ["sync", "test-util"] }
 
 [package.metadata.deb]
+maintainer-scripts = "debian/"
+assets = [
+  ["target/release/orb-supervisor", "/usr/local/bin/", "755"]
```
TheButlah, line 39: I tested it locally - cargo deb is smart enough to rewrite the path to use the target platform as well as profile type.

oldgalileo (approved):

PR #185, opened 2024-08-13
orb-qr-link: Signal user-centric signups

andronat:
@marioradonic @danielle-tfh I think this can help us actually transition to full app-centric easier. Let me know of your thoughts.

### DIFF
--- a/orb-qr-link/src/user_data.rs
+++ b/orb-qr-link/src/user_data.rs
@@ -16,6 +16,9 @@ pub struct UserData {
     /// Personal Custody Package version.
     #[serde(default = "pcp_version_default")]
     pub pcp_version: u16,
+    /// Whether the app should perform a app-centric signup.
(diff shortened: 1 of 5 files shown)


danielle-tfh (approved):

orb-qr-link/src/user_data.rs
```diff
@@ -16,6 +16,9 @@ pub struct UserData {
     /// Personal Custody Package version.
     #[serde(default = "pcp_version_default")]
     pub pcp_version: u16,
+    /// Whether the app should perform a app-centric signup.
+    #[serde(default = "default_false")]
```
valff, line 20: This can be simplified, because the default value for `bool` is `false`. `default_false` function can be removed.

```suggestion
    #[serde(default)]
```

valff (approved):

PR #209, opened 2024-09-04
nix: rewrote everything and added nixos

TheButlah:


fouge:
fixes ORBP-209

fouge (approved):
works on macos

PR #218, opened 2024-09-09
orb-ui: progress circle during biometric capture

fouge:
instead of the slider on Pearl

fixes O-2797


orb-ui/src/engine/animations/slider.rs
```diff
@@ -57,6 +59,7 @@ impl<const N: usize> Slider<N> {
     }
 
     /// Enable pulsing
+    #[allow(dead_code)]
```
TheButlah, line 62: The new rust version adds the `expect` lint, which will error if the lint is not triggered. We should prefer that lint, so that we get informed when the exception is no longer necessary.

https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#expectlint

TheButlah (approved):

Qbicz (approved):

Qbicz:
newb question: why merge commit, and not rebase on main?

fouge:
we want to keep track of commit signatures and this cannot be done with rebase

TheButlah (approved):

PR #220, opened 2024-09-12
orb-ui: sound files in debian package

fouge:
ensure compatibility & correctness of the sounds with a specific orb-ui binary: we don't rely anymore on a sound package version tracked in orb-os. See also https://github.com/worldcoin/orb-os/pull/507

I tested that it correctly fails on checking the sound files in case something is wrong in the `assets` directory, [see job](https://github.com/worldcoin/orb-software/actions/runs/10831070034).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.wav filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/check-sound-files.yaml
+++ b/.github/workflows/check-sound-files.yaml
@@ -1,0 +1,22 @@
+name: Check Sound Files
+
+on:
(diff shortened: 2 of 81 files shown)


TheButlah (changes requested):
Lets use git lfs to store .wav files. This will avoid bloating the git history.

fouge:
@TheButlah this is what I did 🤔

TheButlah:
> @TheButlah this is what I did 🤔
The .gitattributes path was not right, so they didn't get tracked.

I've force pushed and fixed it, will merge in a sec.

TheButlah (approved):

PR #222, opened 2024-09-13
orb-ui: Orb UI self serve ORBP-245

sri9311:
RGB color calibrations as well as a couple new events

### DIFF
--- a/orb-ui/src/engine/diamond.rs
+++ b/orb-ui/src/engine/diamond.rs
@@ -261,6 +261,17 @@ impl EventHandler for Runner<DIAMOND_RING_LED_COUNT, DIAMOND_CENTER_LED_COUNT> {
                 self.operator_action
                     .trigger(1.0, Argb::OFF, true, false, true);
             }
+            Event::SelfServeWait => {
+                self.stop_ring(LEVEL_FOREGROUND, true);
+                self.stop_ring(LEVEL_NOTICE, false);
+                self.set_ring(
+                    LEVEL_FOREGROUND,
+                    animations::Static::<DIAMOND_RING_LED_COUNT>::new(
+                        Argb::DIAMOND_SHROUD_SCAN_USER_AMBER,
+                        None,
+                    ),
+                );
+            }
             Event::SignupStart => {
                 self.capture_sound.reset();
                 self.sound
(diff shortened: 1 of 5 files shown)


fouge:
a few things:
- change the pr title, see the github action, in our case we want the title to be `orb-ui: ...`.
- adding new events is not a small change as orb-core also needs to support the new events, and if possible, we want to make sure we are backward compatible (operator-based + self-serve). Please use the existing events if possible
    - in this case i think that's possible: the biometric capture failure already exists in the current version: there is the timeout event. 
    - I synced with @andronat for the new event `BiometricCaptureStart` because I think it's a necessary one given the current set of events.
    - `SelfServeWait` <=> `Idle`, we might want to ensure we know that we are in self-serve mode though. It's a backend config like the sound volume and language. I think we should do better than just passing the info from orb-core to orb-ui over dbus. Let's discuss that topic with the orb-core ppl.

sri9311:
@fouge the timeout event is specifically `QrScanTimeout`
Is this event used elsewhere? it seems different than the "general" failure that the design team is requesting

fouge:
@sri9311 we want to use the existing `SignupFail` event with the reasons described here
https://github.com/worldcoin/orb-software/blob/9a3eb4ec501468b1eff6587f9ed0938b76faaa36/orb-ui/src/engine/mod.rs#L143

orb-ui/src/engine/diamond.rs
```diff
@@ -652,6 +665,29 @@ impl EventHandler for Runner<DIAMOND_RING_LED_COUNT, DIAMOND_CENTER_LED_COUNT> {
...
                 self.operator_signup_phase.iris_scan_complete();
             }
+            Event::BiometricCaptureFailure => {
+                self.stop_ring(LEVEL_FOREGROUND, true);
+                self.stop_ring(LEVEL_NOTICE, true);
+                self.sound
+                    .queue(sound::Type::Melody(sound::Melody::SoundError))?;
+                // custom alert animation on ring
+                // a bit off for 500ms then on with fade out animation
+                // twice: first faster than the other
```
fouge, line 675: when testing, it animates only once

sri9311:
Closing this out as @fouge has ported over the color changes to his PR here:https://github.com/worldcoin/orb-software/pull/219

PR #224, opened 2024-09-18
ui: bump version

TheButlah:
Since we added the .wav files into the .deb, we should bump the version number.


TheButlah:
Kicked off a -tmp release: https://github.com/worldcoin/orb-software/actions/runs/10926446370

Qbicz (approved):

PR #229, opened 2024-09-20
attest: break dbus into separate crate

TheButlah:
Rebased on #227 

This PR breaks the dbus proxy and interface of orb-attest into a separate crate, and deletes all other copy-pastad zbus::proxy definitions for orb-attest.

This also paves the way to mock orb-attest if we need to in tests.

I intend to repeat this same pattern for all the other dbus interfaces, if reviews are favorable.


orb-supervisor/tests/it/helpers.rs
```diff
@@ -190,8 +189,8 @@ pub async fn start_interfaces(
     dbus_instances: &DbusInstances,
 ) -> zbus::Result<zbus::Connection> {
     let conn = zbus::ConnectionBuilder::address(dbus_instances.system.address())?
-        .name(zbus_systemd::systemd1::ManagerProxy::DESTINATION)?
-        .serve_at(zbus_systemd::systemd1::ManagerProxy::PATH, Manager)?
+        .name(zbus_systemd::systemd1::ManagerProxy::DESTINATION.unwrap())?
```
Qbicz, line 192: if we're about to panic here, would you consider `expect()` with super short explanation? I would have no clue how to fix/debug the panic here without going deeper as I can't find the DESTINATION and PATH values in our code nor the `zbus_systemd` crate.

Qbicz: Ok actually no need for more info here - can be figured out from the code

vmenge (approved):
Very nice stuff! Looking forward to the increase on stability on interface changes :D

PR #232, opened 2024-09-20
utils: write the orb-id in the fuse-banks

SfikasTeo:


Qbicz (commented):
Looks good, is it supposed to be included in Orb's rootfs and run on Orb, as opposed to https://github.com/worldcoin/orb-software/pull/166 which is meant to be run from the outside over SSH?

TheButlah (approved):

PR #235, opened 2024-09-25
orb-slot-ctrl: integration tests and OrbSlotCtrl struct

vmenge:
added integration tests and test utils on `orb-slot-ctrl`
test utils are meant to be used by other crates when testing operations done through `orb-slot-ctrl`.
small refactoring done with the introduction of `EfiVarDb` to acquire `EfiVar`  from arbitrary paths.

tested and working on an Orb :+1:


CODEOWNERS
```diff
@@ -1 +1,2 @@
 .github/* @fouge @TheButlah @vmenge @oldgalileo
+orb-slot-ctrl/* @vmenge @oldgalileo @alekseifedotov
```
TheButlah, line 2: I don't want to add individual codeowners to specific software components. Anyone should be able to review code in this repo. We do not need to further bottleneck our review process.

vmenge: I'm okay with postponing this discussion for now, but I strongly feel a cross-team monorepo does not work well without a _CODEOWNERS_ file. IMO every crate / lib should have a team as clear owner as long as we're not bottlenecked by a single person.

update-verifier/src/lib.rs
```diff
@@ -21,13 +22,14 @@ pub const BUILD_INFO: BuildInfo = make_build_info!();
 pub fn run_health_check() -> eyre::Result<()> {
     // get runtime environment variable to force health check
     let dry_run = std::env::var("UPDATE_VERIFIER_DRY_RUN").is_ok();
+    let orb_slot_ctrl = OrbSlotCtrl::new(&EfiVarDb::from_rootfs("/")?)?;
```
oldgalileo, line 25: I know we already went over this, but I'll admit I'm still skeptical of dependency injection _even in application code_ as the answer to testability.

We looked a bit together, and I'll admit that alternatives besides doing something a bit more hack-y don't immediately jump out at me here. But I think in general testing should be _**in service of**_ application code.

oldgalileo: I thought more about this, and I think the issue I have with this testing is that it's testing `File::open` and `File::read`. That test, applied to the `Fixture` here, does not match the real-world scenario of reading and writing to the `efifvars`.

orb-slot-ctrl/src/main.rs
```diff
@@ -1,204 +1,14 @@
...
-        Commands::GitDescribe => {
-            println!("{}", BUILD_INFO.git.describe);
-        }
-    }
-    Ok(())
+
+    let db = EfiVarDb::from_rootfs("/")?;
+    let orb_slot_ctrl = OrbSlotCtrl::new(&db)?;
+
+    program::run(&orb_slot_ctrl, cli)
```
oldgalileo, line 13: Why do we move `program` into a new file?

orb-slot-ctrl/src/efivar/bootchain.rs
```diff
@@ -5,17 +5,30 @@
...
 //! Bits of interest are found in byte 4 for all efivars.
 
-use super::{is_valid_buffer, EfiVar, SLOT_A, SLOT_B};
+use super::{is_valid_buffer, EfiVar, EfiVarDb, EfiVarDbErr, SLOT_A, SLOT_B};
 use crate::Error;
 
-pub const PATH_CURRENT: &str =
-    "BootChainFwCurrent-781e084c-a330-417c-b678-38e696380cb9";
-pub const PATH_NEXT: &str = "BootChainFwNext-781e084c-a330-417c-b678-38e696380cb9";
+const PATH_CURRENT: &str = "BootChainFwCurrent-781e084c-a330-417c-b678-38e696380cb9";
```
oldgalileo, line 11: I know that you only changed their visibility, but it would be awesome to get a page reference to this in the EFI spec or wherever these values came from. Magic values with no documentation = sad face.

orb-slot-ctrl/src/main.rs
vmenge: We can't import things from `main.rs`  on integration tests :(

oldgalileo (approved):
@vmenge and I talked in person.

I stand by what I said before, that testing `fs::write` is **not** the aim of these tests. There's a point here that we do want to enable verification of the state after a series of operations (i.e. we have the `UpdateDone` rootfs status at the beginning of an integration-tested update-verifier run, and then we validate that it's `Normal` at the end). Remains to be seen how to appropriately implement that

orb-slot-ctrl/src/efivar/bootchain.rs
vmenge: agreed! i'll chase down where these come from and document them later :+1:

orb-slot-ctrl/src/test_utils.rs
```diff
@@ -0,0 +1,78 @@
...
+        fs::create_dir_all(&db_path).unwrap();
+
+        let db = EfiVarDb::from_rootfs(&tempdir).unwrap();
+        let bootchain = BootChainEfiVars::new(&db).unwrap();
+        let rootfs = RootfsEfiVars::new(&db).unwrap();
+        let slot_ctrl = OrbSlotCtrl::new(&db).unwrap();
+
+        bootchain
+            .current
+            .create_and_write(&[0x07, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
```
TheButlah, line 36: Explain these magic values.

orb-slot-ctrl/tests/orb_slot_ctrl.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[test]
+fn it_gets_current_slot() {
+    let fx = Fixture::new();
+    let slot = fx.slot_ctrl.get_current_slot().unwrap();
+    assert_eq!(slot, Slot::A)
+}
+
+#[test]
+fn it_gets_inactive_slot() {
```
TheButlah, line 12: Question: Do you have an opinion on whether test functions should start their names with `test_` or not? In some cases I think it could disambiguate cases where we forgot to annotate with #[test]. But IDK.

orb-slot-ctrl/tests/orb_slot_ctrl.rs
```diff
@@ -0,0 +1,77 @@
...
+    let fx = Fixture::new();
+    let slot = fx.slot_ctrl.get_inactive_slot().unwrap();
+    assert_eq!(slot, Slot::B)
+}
+
+#[test]
+fn it_gets_and_sets_next_boot_slot() {
+    let fx = Fixture::new();
+    let slot = fx.slot_ctrl.get_next_boot_slot().unwrap();
```
TheButlah, line 21: Why would it default to the next slot being B? I would assume A is the current, and the next slot remains A.
Is this a product of the mystery writes in the Fixture code?

update-verifier/src/main.rs
```diff
@@ -25,25 +26,20 @@ fn clap_v3_styles() -> Styles {
...
-        telemetry::start::<ExecContext, _>(LevelFilter::INFO, std::io::stdout)
-    {
-        error!(
-            ?error,
-            "update verifier encountered error while starting telemetry"
-        );
-        std::process::exit(1);
-    }
+fn main() -> eyre::Result<()> {
+    run().inspect_err(|error| error!(?error, "failed to run update-verifier"))
```
TheButlah, line 30: Why do it this way?


Also, main should be using `color_eyre::Result` along with a `color_eyre::install()?`

TheButlah (changes requested):
Overall looks great! Thanks @vmenge :3

Just a few small changes and we will be good.

update-verifier/src/main.rs
vmenge: > Why do it this way?
No need to manually call process::exit like was being done befoe

> Also, main should be using color_eyre::Result along with a color_eyre::install()?
Sure

orb-slot-ctrl/src/test_utils.rs
vmenge: I got them all from the unit tests from `rootfs.rs`  and `bootchain.rs`, and commits were done by you and Mark Eder, so you probably can explain them better than me :sweat_smile:

orb-slot-ctrl/tests/orb_slot_ctrl.rs
vmenge: Nah, I don't think its necessary... starting tests with `it_` is a convention in some places as well, but I don't really think it matters since we have to annotate with `#[test]`.

orb-slot-ctrl/Cargo.toml
```diff
@@ -18,8 +18,15 @@ eyre.workspace = true
...
 [build-dependencies]
 orb-build-info = { path = "../build-info", features = ["build-script"] }
 
 [package.metadata.deb]
+
+[features]
+test = []
+
+[package.metadata.orb]
+unsupported_targets = ["aarch64-apple-darwin", "x86_64-apple-darwin"]
```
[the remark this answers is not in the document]
vmenge: It's not due to tests though, it's due to the implementation which uses `libc` (you do have an upcoming change in the other PR). So the bin for it wouldn't even run in macOS.

I'll leave this as is for now as you can remove it once you've merged the removal of `libc`. I am not a fan of annotating every single test we have because the macOS pipeline can't run it tbh :/

PR #236, opened 2024-09-26
zbus-proxies: initial commit and add login1

TheButlah:
This crate will become the new home of zbus proxies for third party services. The intention is to then use the `login1` proxy in the update agent, and remove `zbus_systemd`, since the maintainer has repeatedly stated that [our changes won't be upstreamed](https://github.com/lucab/zbus_systemd/issues/41). For something so trivial, we are better off rewriting it ourselves, and consolidating all of this in one spot.

We might also want other third party services like wpa_supplicant to live in here in the future.

The PR adds .xml files retrieved via busctl introspect, as well as a cli tool to convert from the xml files to the zbus code. It then includes the generated results after tweaking them manually and documenting the tweaks.


TheButlah:
I'm closing this PR. zbus-xmlgen doesn't generate all of the additional proxy structs for systemd1. It seems to only generate the main toplevel manager. I'm also dissatisfies that I have to include zbus-xmlgen as a binary via nix instead of using it as a library.

Given that zbus_systemd is fine to use as long as we only use the async interfaces, I've decided that forking it as @oldgalileo originally did, or simply vendoring the proxy interface ourselves when we can't use async, is fine.

PR #237, opened 2024-09-27
orb-ui: queue sound with an initial delay

fouge:
delay is awaited until elapsed before playing the sound.


orb-ui/src/sound/mod.rs
```diff
@@ -37,7 +38,8 @@ pub trait Player: fmt::Debug + Send {
 
     /// Queues a sound to be played.
     /// Helper method for `build` and `push`.
-    fn queue(&mut self, sound_type: Type) -> Result<()>;
+    /// Optionally delays the sound.
+    fn queue(&mut self, sound_type: Type, delay: Option<Duration>) -> Result<()>;
```
TheButlah, line 42: Should just be a `Duration`. No delay means Duration::ZERO

TheButlah (approved):

PR #240, opened 2024-09-27
orb-ui: control the gimbal

fouge:
will allow usage of the gimbal in the
"show-car" simulation

### DIFF
--- a/orb-ui/src/engine/mod.rs
+++ b/orb-ui/src/engine/mod.rs
@@ -344,6 +344,12 @@ event_enum! {
         /// Plays boot-up complete sound for testing
         #[event_enum(method = sound_test)]
         SoundTest,
+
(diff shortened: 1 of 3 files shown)


orb-ui/src/engine/mod.rs
```diff
@@ -344,6 +344,11 @@ event_enum! {
         /// Plays boot-up complete sound for testing
         #[event_enum(method = sound_test)]
         SoundTest,
+
+        #[event_enum(method = gimbal)]
+        Gimbal {
```
TheButlah, line 349: need to document - what are the units, and are they absolute values or deltas?

TheButlah (changes requested):

TheButlah (approved):

PR #244, opened 2024-09-27
orb-ui: use Duration::ZERO instead of Option<Duration>

fouge:
no need for an option where the default can be used

fixes ORBP-266

### DIFF
--- a/orb-ui/sound/src/queue.rs
+++ b/orb-ui/sound/src/queue.rs
@@ -38,7 +38,7 @@ pub struct SoundBuilder<'a> {
     reader: Option<Box<dyn Reader>>,
     cancel_all: bool,
     priority: u8,
-    delay: Option<Duration>,
+    initial_delay: Duration,
     max_delay: Duration,
     volume: f64,
 }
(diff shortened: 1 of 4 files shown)


Qbicz:
The version with the Option and default argument looked less verbose to me in every call ;)
Do you ever expect non-zero delay duration? If it's not common (I don't see it in any of the cases when sound is played) we could do something like this - have a default value in the `queue()` and then you don't have to pass down the Option to `pub fn delay(mut self, delay: Option<Duration>)` but will have `pub fn delay(mut self, delay: Duration)` instead
```
    fn queue(&mut self, sound_type: Type, delay: Option<Duration>) -> Result<()> {
        let volume = self.volume;
        let initial_delay = delay.unwrap_or(Duration::ZERO);
        self.build(sound_type)?.volume(volume).delay(initial_delay).push()?;
        Ok(())
    }
```
but it's just subjective, feel free to go with this one too

Qbicz:
Ok I've seen the other PR introducing the delay argument. Makes sense now

Qbicz (approved):

TheButlah (approved):

TheButlah (approved):

PR #246, opened 2024-09-30
slot-ctrl: remove two lines of unsafe

TheButlah:


fouge (commented):
is it really worth it? 🤔 
it delegates the unsafe code to a third party crate but should we trust libc or rustix more?

vmenge:
> is it really worth it? 🤔 it delegates the unsafe code to a third party crate but should we trust libc or rustix more?

Using `libc` like this causes issues on macOS :/

vmenge (approved):
LGTM! Did you test it on an orb?

TheButlah:
> is it really worth it?

Yes, long term I want to try to use rustix as much as possible.

I will eventually rewrite the Ioctl stuff using it. It provides good tools for Ioctl calls. See for example my PR to nusb: https://github.com/kevinmehall/nusb/pull/65

TheButlah:
> Did you test it on an orb?

No, because the thing is the same function. But I can do that before merging.

fouge:
@TheButlah let's resolve the conflicts and merge

PR #252, opened 2024-10-08
orb-mcu-util: hw rev: overwrite file if content is different

fouge:
to prevent useless writes to flash.
++ use tokio::fs instead of std::fs


TheButlah (approved):
@fouge loves blocking in async :P

fouge:
@TheButlah i noticed the issue while trying to make this command smarter 😬

PR #256, opened 2024-10-09
zbus-proxies: add login1 proxy

TheButlah:
Reviving #236 because fuck it we ball


TheButlah:
forcing merge, its needed for the update agent.

PR #258, opened 2024-10-10
telemetry: introduce orb-telemetry crate and use it in orb-ui

TheButlah:
we should in general be using journald for our logs, not stdout. Also, we should use stderr instead of stdout.

Every one of our crates sets up logging in a different way, but they all do the same thing. I wanted to introduce a single spot that lets us configure our telemetry. If it goes well in orb-ui, we can use it in all the rest too

### DIFF
--- a/telemetry/src/lib.rs
+++ b/telemetry/src/lib.rs
@@ -1,0 +1,84 @@
+use std::io::IsTerminal as _;
+
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{
+    layer::SubscriberExt as _, util::SubscriberInitExt as _, EnvFilter,
+};
+
+#[derive(Debug)]
+pub struct TelemetryConfig {
+    syslog_identifier: Option<String>,
+    global_filter: EnvFilter,
+}
+
+impl TelemetryConfig {
+    /// Provides all required arguments for telemetry configuration.
+    /// - `log_identifier` will be used for journald, if appropriate.
+    #[expect(clippy::new_without_default, reason = "may add required args later")]
+    pub fn new() -> Self {
+        Self {
+            syslog_identifier: None,
+            global_filter: EnvFilter::builder()
+                .with_default_directive(LevelFilter::INFO.into())
+                .from_env_lossy(),
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


telemetry/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+        Self {
+            global_filter: filter,
+            ..self
+        }
+    }
+
+    /// Initializes the telemetry config. Call this only once, at the beginning of the
+    /// program.
+    ///
+    /// Calling this more than once or when another tracing subscriber is registered
```
fouge, line 49: should we prevent the panic? returns an error instead?

telemetry/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// will cause a panic.
```
fouge, line 50: should we prevent the panic? return an error instead?

fouge (commented):
I tested and the filtering doesn't seem to work as expected

I have `RUST_LOG=debug` in my environment, but the same applies if the `default_directive` uses `LevelFilter::DEBUG`:
- when filtering out `info` and above levels with `journalctl`, i still have debug logs (ie `tracing::debug!()`)
- these disappear only when I filter out `warning` and above levels 🤔 
 
```
(venv) (os-b5c3a10) worldcoin@id-faff46c4:/mnt/scratch$ jl -fu worldcoin-ui --priority=info
-- Logs begin at Tue 2023-11-21 21:10:22 UTC. --
Oct 23 08:43:17 localhost worldcoin-ui[2338]: UI event: {"BatteryIsCharging":{"is_charging":true}}
[...]
```

corresponding line
```
tracing::debug!("UI event: {}", serde_json::to_string(event)?.as_str());
```

telemetry/src/lib.rs
TheButlah: I think the panic is intended. If we called it more than once we should abort.

fouge:
should we document this somewhere? 🤔

TheButlah:
Its already documented by tracing_journald.

fouge (approved):

PR #259, opened 2024-10-11
seek-camera: bump seek sdk version

TheButlah:
@AlexKaravaev this bumps the seek thermal sdk version to the latest. Maybe it will help fix the CID issues we have been noticing at jabil. Can you can test this on the orb that is having the CID issues?


TheButlah:
since it didn't help, we are closing.

PR #263, opened 2024-10-15
orb-ui: pulsates ring with a breathe/wave animation.

fouge:
with 3-second period

temporary workaround?


karelispanagiotis (approved):
I could only test it with `orb-ui simulation` mode since my orb's IR camera is broken.
@fouge performed a full test.
Approving!

PR #265, opened 2024-10-16
ui: diamond: tweaked a few animations

fouge:
qr capture animations, smoother transitions, faster error alert


Qbicz (approved):
Looks good!

PR #268, opened 2024-10-17
orb-mcu-util: control gimbal

fouge:
allows to trigger auto-home and set gimbal position for testing.

fixes O-2915

### DIFF
--- a/mcu-util/src/orb/mod.rs
+++ b/mcu-util/src/orb/mod.rs
@@ -77,18 +77,18 @@ impl Orb {
         ))
     }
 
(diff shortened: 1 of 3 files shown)


mcu-util/src/orb/mod.rs
```diff
@@ -84,6 +84,10 @@ impl Orb {
         }
     }
 
+    pub fn borrow_mut_main_board(&mut self) -> &mut MainBoard {
```
TheButlah, line 87: could call it `main_board_mut(&mut self)` and `main_board(&self)`? This is the convention in rust.

TheButlah (approved):

TheButlah (approved):

TheButlah (approved):

PR #269, opened 2024-10-17
hil: add option for reusing existing rts

AlexKaravaev:
s3 urls are unique for given commit, we might want to reuse them for gh workflows, but currently since we manipulate filename inside orb-hil we cannot supply correct rts-path (We can in theory but that's ugly)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1069,7 +1069,7 @@ dependencies = [
  "bitflags 2.4.2",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.12.1",
  "lazy_static",
  "lazycell",
  "log",
@@ -2881,7 +2881,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.4.10",
+ "socket2 0.5.5",
  "tokio",
  "tower-service",
  "tracing",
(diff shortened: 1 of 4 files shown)


TheButlah (changes requested):
We should not add functionality that is brittle. Unless we perform a checksum validation on the file, and retrieve the canonical sha from s3, I don't think we should reuse any files.

AlexKaravaev:
yep, will do so. It should be fairly easy.

AlexKaravaev:
It turned out to be trickier than I expected:

* AWS requires the user to set a checksum when uploading; otherwise, it splits the file into multiple parts, and "Amazon S3 concatenates the bytes for the MD5 digests together and then calculates the MD5 digest of these concatenated values." (See [AWS Documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#checking-object-integrity-etag-and-md5)).

* The number of parts seems constant in our case but essentially depends on the version of the AWS S3 CLI. For me, the chunk size was 8192 KB.

* I don’t want to hard-code these numbers, but I haven’t found a way to retrieve them reliably at runtime. I suggest we close this for now and revisit it later. The optimization here is that we avoid downloading the RTS for the same commit, but in practice, it probably won’t happen often that we need to download RTS in CI for the same commit. Plus, we only save about 6 minutes with this.

PR #277, opened 2024-10-21
orb-mcu-util: add --can-fd option

fouge:
for all commands
usage: orb-mcu-util --can-fd args

fixes O-2891


TheButlah (approved):
looks good. but why did we remove the heartbeat check? That was useful in the past

PR #281, opened 2024-10-25
feat(hil): Add button control cmd and timeout for cmd

AlexKaravaev:


fouge (approved):
apart from delegating the parsing of the timeout, the rest looks good to me :) 
(not tested though)

AlexKaravaev:
@TheButlah what do we do about security complaining? https://github.com/zeta12ti/parse_duration/issues/21 this is not relevant at all for cli command use-case probably

AlexKaravaev:
nvm, will switch to https://docs.rs/humantime/latest/humantime/

TheButlah (approved):

PR #282, opened 2024-10-28
test(sound): check sound files in rust tests

fouge:
instead of invoking soxi
check that all files are present in the assets
check bits per sample (16 bits)
SOUNDS_DIR can be overwritten at runtime with an env variable tests are run with all other rust tests


.github/workflows/rust-ci.yaml
```diff
@@ -148,8 +149,8 @@ jobs:
           cat ${GITHUB_ENV}
       - name: Cargo Test
         run: |
-          nix develop -c \
-            cargo test --all --all-features --all-targets $EXCLUDES
+          SOUNDS_DIR=$(pwd)/ui/sound/assets nix develop -c \
```
fouge, line 152: i don't like setting this variable here but i didn't find a better approach :/

TheButlah: lets use a config struct rather than this global env var

TheButlah (changes requested):

TheButlah (approved):

PR #284, opened 2024-10-31
feat(agentwire): Initial commit, extracted from `orb-core`

valff:
Open-source `agentwire` framework.


valff:
@TheButlah this is a linux-only crate. Can we skip the macos check?

TheButlah:
Yes, we can opt-out of macos support the following way:
https://github.com/worldcoin/orb-software/blob/755c9bca2f87b26846b201079d56a6b41c06a837/slot-ctrl/Cargo.toml#L32

PR #287, opened 2024-11-13
scripts: supporting remote re-provisioning

sfikastheo:
Fixes: ORBS-146

### DIFF
--- a/scripts/remote-provision.sh
+++ b/scripts/remote-provision.sh
@@ -1,0 +1,144 @@
+#!/usr/bin/env bash
+
+set -o errexit   # abort on nonzero exit status
+set -o errtrace  # pass ERR trap down to functions, substitutions, etc
+set -o nounset   # abort on unbound variable
+set -o pipefail  # don’t hide errors within pipes
+
+[[ "${BASH_VERSINFO:-0}" -ge 4 ]] || { echo "Bash version 4 or higher is required."; exit 1; }
+
+usage() {
+    echo "Usage $0 [OPTIONS] <teleport-tunnel|ipv4>
+
+    Options:
+        -h, --help          Print help
+        -r, --reprovision   Re-provisioning the device and skips fetching of attestation certificate
+        -p, --passphrase    Expected to be used with IPv4 (required for dev images with ssh access)
+        -t, --plug-trust    Path for plug_and_trust.tar.gz on the host (required for prod images).
+                            Check the worldcoin/plug-and-trust/releases for the appropriate version
+                            based on the device's software release
+        -y, --assumeyes     automatically answer yes for all questions
+
+    Example:
+        $0 -r -t ~/Downloads/plug_and_trust.tar.gz <teleport-tunnel>"
+}
+
+provision_device() {
+    local remote="${1}"
+    local reprovision="${2}"
+    local ssh_prefix="${3}"
+    local plug_trust="${4}"
+    local interactive="${5}"
+
+    local user="worldcoin"
+    local se_dir="/usr/persistent/se"
+    local key_dir="${se_dir}/keystore"
+    if [[ "${ssh_prefix}" == "tsh" ]]; then
+        user="root"
+    fi
+
+    # If /se/keystore is not present, provisioning process was never executed, or keystore was wiped
+    # In this case, reprovisioniong is not allowed for fear of wiping the attestation certificate
+    if [[ ${reprovision} ]]; then
+        if [[ ${interactive} ]]; then
+            read -p "Reprovisioning will wipe all provisioning material, continue? [y/N] " -n 1 -r
+            echo
+            if [[ ! ${REPLY} =~ ^[Yy]$ ]]; then
+                echo "Reprovisioning aborted"
+                exit 0
+            fi
+        fi
+        ${ssh_prefix} ssh "${user}@${remote}" bash --noprofile --norc <<EOF
+        set -euo
+        if ! [[ -d ${key_dir} ]]; then
+            echo '${key_dir} does not exist, re-provisioning blocked'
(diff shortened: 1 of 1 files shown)


scripts/remote-provision.sh
```diff
@@ -0,0 +1,119 @@
...
+        ssh_prefix="sshpass -p "${passphrase}""
+    fi
+
+    local plug_trust_target="/service_mode"
+    if ${ssh_prefix} ssh "worldcoin@${remote}" "! [[ -d ${plug_trust_target} ]]" > /dev/null 2>&1; then
+        # NOTE: The following case, should only be met in production images, where tsh
+        # is the only available option for remote access. SOON will be deprecated
+        if [[ -n "${plug_trust}" ]]; then
+            ${ssh_prefix} scp "${plug_trust}" "root@${remote}:/tmp/plug_and_trust.tar.gz"
+            echo "Started proviosining"
```
AlexKaravaev, line 106: This should be fixed, it's stucks here

AlexKaravaev (changes requested):

scripts/remote-provision.sh
```diff
@@ -0,0 +1,128 @@
...
+set -o nounset   # abort on unbound variable
+set -o pipefail  # don’t hide errors within pipes
+
+[[ "${BASH_VERSINFO:-0}" -ge 4 ]] || { echo "Bash version 4 or higher is required."; exit 1; }
+
+usage() {
+    echo "Usage $0 [OPTIONS] <teleport-tunnel|ipv4>
+
+    Options:
+        -r, --reprovision   Re-provisioning the device and skips fetching of attestation certificate
```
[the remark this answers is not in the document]
sfikastheo: In case the `usr/persistent/se/keystore` does not exist, the script should exit. I think that this is enough of a test to ensure that the attestation certificate is not going to get wiped on unprovisioned orbs.

scripts/remote-provision.sh
```diff
@@ -0,0 +1,129 @@
...
+    local ssh_prefix="tsh"
+    local positional_args=()
+
+    while [[ $# -gt 0 ]]; do
+        arg="${1}"; shift
+        case ${arg} in
+            -h | --help)
+                usage; exit 0 ;;
+            -r | --reprovision)
+                short=true ;;
```
AlexKaravaev, line 78: why do we call "reprovision" as "short" ? for me it's a bit misleading?

scripts/remote-provision.sh
```diff
@@ -0,0 +1,129 @@
...
+    # If /se/keystore is not present, provisioning process was never executed, or the folder was deleted by the user
+    # In this case, reprovisioniong (--short flag) is not allowed for fear of wiping the attestation certificate
+    if [[ ${short} ]]; then
+        ${ssh_prefix} ssh "${user}@${remote}" << EOF
+        set -euo
+        sudo su || true
+        [[ -d ${key_dir} ]] || echo "${key_dir} not exist, re-provisioning blocked" && exit 1
+        mount -o remount,exec /tmp
+        systemctl stop nv-tee-supplicant.service
+        cp /${key_dir}/f0000013.cert /usr/persistent/ || true
```
AlexKaravaev, line 45: do we really need to shadow cp exit code?
I am thinking of smth very stupid:
```
someone creates f0000013.cert *directory* at key_dir
cp command returns true despite that cp will actually fail
we end up wiping out cert
```

Maybe we just remove `|| true` ?

scripts/remote-provision.sh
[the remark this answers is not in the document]
AlexKaravaev: I don't see connection. `/usr/persistent/se/keystore` even if this dir exists `-r` seems very intrusive, we will preserve att.token, but new certs will be generated that later needs to be uploaded.

I would still prompt user and add `-f/force` flag for disabling prompt.

scripts/remote-provision.sh
AlexKaravaev: I mean I do see why we call it short, but I don't agree with it. I think variable should be called `reprovision` for clarity.

AlexKaravaev (changes requested):

TheButlah (approved):
Approving to unblock.

scripts/remote-provision.sh
sfikastheo: to be inline with the arguments in `plug-and-trust` repo. 
Changed to avoid confusion

scripts/remote-provision.sh
sfikastheo: you can skip the shadowing, and add explicit checks. The situation you described would require maliciousness from a developer. I do not see a reason to be so overly protective.

The reason the `cp`  return value is shadowed, is because on most dev orbs, the cert is lost in the first place. if the consensus is to be more robust, i can try checking `os-release`

scripts/remote-provision.sh
```diff
@@ -0,0 +1,141 @@
...
+[[ "${BASH_VERSINFO:-0}" -ge 4 ]] || { echo "Bash version 4 or higher is required."; exit 1; }
+
+usage() {
+    echo "Usage $0 [OPTIONS] <teleport-tunnel|ipv4>
+
+    Options:
+        -r, --reprovision   Re-provisioning the device and skips fetching of attestation certificate
+        -p, --passphrase    Expected to be used with IPv4 (required for dev images with ssh access)
+        -t, --plug-trust    Path for plug_and_trust.tar.gz on the host (required for prod images).
+                            Check the worldcoin/plug-and-trust/releases for the appropriate version
```
AlexKaravaev, line 17: last nit is to add -y option to usage and we are good to go!

AlexKaravaev (approved):

PR #288, opened 2024-11-14
scripts: nit support 'short' certificate uploading

sfikastheo:


AlexKaravaev:
We can create other script for that but feedback from batch fixing orbs is that copy and upload should not be 2 different scripts. It's handy to just write one. Or (ideally) all of them 3 should be placed later in one script

sfikastheo:
@AlexKaravaev Lets merge this, and i will just push a quick-fix to have a just `fetch` option

scripts/upload-certs.sh
```diff
@@ -92,26 +97,24 @@ main() {
...
     cf_token="$(get_cloudflared_token "${domain}")"
 
-    # Post certificate
-    curl --fail --location \
-        -H "Authorization: Bearer ${bearer}" \
-        -H "cf-access-token: ${cf_token}" \
-        -X POST "https://${domain}/api/v1/certificate" \
-        -d '{ "orbId": "'"${orb_id}"'", "certificate": "'"${certificate}"'" }'
+    # Post attestation certificate
+    if [[ ! "${short}" ]]; then
```
AlexKaravaev, line 105: Another usage feedback that I had is that a lot of times dir name is the orb-id, so we don't need to provide that explicitely. We can use that in next script that combines both copy/upload.

AlexKaravaev (approved):

PR #290, opened 2024-11-14
fix: remove clang override for stdenv and switch runner to unstable

TheButlah:


TheButlah:
I think it works, but to really test it it requires me to have admin perms for orb-os so that I can re-register the runner.

TheButlah:
It works, I tested.

AlexKaravaev (approved):

PR #292, opened 2024-11-18
feat(jwk-util): Initial commit

TheButlah:
Implements a CLI tool to convert from pcks8 encoded private keys to JWK public keys.


ik-work-gh:
Added bug bounty disclaimer & an assert after review w/ Ryan!

ik-work-gh (approved):

PR #295, opened 2024-11-19
fix(update-agent): increase ack timeout for mcu

TheButlah:
Backport changes from the now-archived update-agent repo.


fouge (approved):
🙏

PR #307, opened 2024-12-03
update-agent: refactor

alekseifedotov:
Update agent deviates from conventional rust coding pattern. For example it uses `&String` instead of `&str` and `&Option<String>` instead `Option<&String>`. 

I've also removed some structures like `version_map::ComponentInfo` and `version::ComponentInfo` (isn't it fun to have two different structures with the same name?)

--

I would appreciate comments about rust style/code quality and what else I could have refactored.

### DIFF
--- a/update-agent/core/src/version_map.rs
+++ b/update-agent/core/src/version_map.rs
@@ -115,10 +115,10 @@ pub struct ComponentIter<'a> {
 }
 
 impl<'a> Iterator for ComponentIter<'a> {
-    type Item = (&'a String, &'a ComponentInfo);
+    type Item = (&'a str, &'a ComponentInfo);
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.inner.next()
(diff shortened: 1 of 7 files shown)


update-agent/core/src/version_map.rs
```diff
@@ -137,7 +121,7 @@ impl VersionMap {
         }
     }
 
-    pub fn get_component(&self, name: &str) -> Option<&ComponentInfo> {
+    pub fn get_component(&self, name: &str) -> Option<&SlotVersion> {
```
TheButlah, line 124: perhaps it should be renamed to slot_version()? for two reasons:
* the type is SlotVersion, not ComponentInfo anymore
* rust getters [don't usually](https://rust-lang.github.io/api-guidelines/naming.html#c-getter) use the get_ prefix.

alekseifedotov: Thank you for the review. I've renamed it in a separate commit.

TheButlah (approved):

PR #310, opened 2024-12-06
Add integration test for update agent

alekseifedotov:
In this PR I'm adding integration test for update agent. 

**How it works**

The script spawns a podman container with update-agent inside. That docker container partially reproduces orb environment. It mocks: 
 - `efivars`
 - `/usr/persistent/`
 - `/dev/mmcblk0`.
 - `/etc/orb_update_agent.conf`
- `claim.json`

In claim.json it contains links on local files to get the update.
After update-agent is done, the script compares the resulting partitions with the expected ones.

Future steps
 

- [ ] add [localstack](https://www.localstack.cloud/) to mock S3 buckets
- [ ] add dbus in the container to mock dbus communication
- [ ] add mcu emulation
- [ ] add capsule update  mock
---
I'll make it a draft for now.
---

**Requirements:** 
- [nu shell](https://www.nushell.sh/)
- [libguestfs](https://libguestfs.org/guestfish.1.html)
- [podman](https://podman.io/)


vmenge (approved):
this is awesome and I think a great start for us in feeling safe with `update-agent` changes :partying_face: 

what I'd ask here is to as a next step (on a separate PR) move this over to something that can be used with both `cargo watch` and `cargo test` for a quick feedback loop when iterating locally :pray:

alekseifedotov:
>what I'd ask here is to as a next step (on a separate PR) move this over to something that can be used with both cargo watch and cargo test for a quick feedback loop when iterating locally 🙏

The test itself takes a couple of minutes, so I don't think it is suitable to run in `cargo watch`

PR #311, opened 2024-12-07
feat(thermal-cam-ctrl): add cleanup command

TheButlah:


thermal-cam-ctrl/src/cleanup.rs
```diff
@@ -0,0 +1,60 @@
...
+    for entry in calib_dir
+        .read_dir()
+        .wrap_err("failed to access SEEKTHERMAL_ROOT/.seekthermal/cal")?
+    {
+        let entry = entry?;
+        if cid_to_keep.as_str() == entry.file_name() {
+            continue; // skips matching dir
+        }
+        tracing::info!("removing {}", entry.path().display());
+        //std::fs::remove_dir_all(entry.path()).wrap_err_with(|| {
```
AlexKaravaev, line 55: Probably forgot to uncomment ?

AlexKaravaev (changes requested):

thermal-cam-ctrl/src/cleanup.rs
TheButlah: whoops yeah. Pushed a commit that fixes this.

TheButlah:
Yes we will still need to call the command, in the systemd one shot

PR #312, opened 2024-12-07
chore(thermal-cam-ctrl): move to /usr/local/bin

TheButlah:


AlexKaravaev (approved):

AlexKaravaev:
We should also change service file definitions in orb-os

PR #317, opened 2024-12-10
feat(qr-link): Age verification token field in UserData struct

bsayed:
Added new boolean field in the UserData struct to enable/disable age verification on the orb.

### DIFF
--- a/qr-link/src/user_data.rs
+++ b/qr-link/src/user_data.rs
@@ -24,9 +24,8 @@ pub struct UserData {
     pub user_centric_signup: bool,
     /// A unique UUID that the Orb will use to send messages to the app through Orb Relay.
     pub orb_relay_app_id: Option<String>,
-    /// Whether the Orb should perform the age verification.
-    #[serde(default = "default_false")]
-    pub bypass_age_verification: bool,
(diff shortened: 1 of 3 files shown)


qr-link/src/user_data.rs
```diff
@@ -24,6 +24,8 @@ pub struct UserData {
     pub user_centric_signup: bool,
     /// A unique UUID that the Orb will use to send messages to the app through Orb Relay.
     pub orb_relay_app_id: Option<String>,
+    /// Whether the Orb should perform the age verification. If the token exists we skip the age verification.
+    pub bypass_age_verification_token: Option<String>,
```
karelispanagiotis, line 28: rename to `bypass_age_verification` ?

karelispanagiotis, line 28: why is this a `String`?
if it should be `bool`, what about a rename to `bypass_age_verification`

karelispanagiotis: I think I got my answer from `...request from POP team`.

karelispanagiotis (approved):

andronat (approved):

PR #331, opened 2024-12-17
fix(remote-prov): No sudo

sfikastheo:


karelispanagiotis (approved):
checked that it works with `tsh`
haven't tried with ssh.

PR #340, opened 2024-12-25
refactor: FM-321 move orb relay client

paulquinn00:
Moving `orb-relay-client` to `orb-software` to make it accessible. Specifically to allow `fleet-cmdr` to utilize it.

Made a patch from `priv-orb-core/orb-relay-client` and applied it to `orb-software/relay-client`.
(Follow-up PR to `priv-orb-core` to depend on `orb-software/relay-client` will come.)

PR #341, opened 2024-12-27
Fm 321 update relay client

paulquinn00:
- set to use rust 1.82
- update to latest relay-messages which requires rust 1.82
- updates to use Relay namespacing


TheButlah (changes requested):
lets revert the toml formatting changes. Then we can re-review once #340 is merged.

PR #345, opened 2025-01-02
fix(upload-certs): vol 3

sfikastheo:
Same mistake as https://github.com/worldcoin/orb-software/pull/335/files
Skill issue :(


AlexKaravaev:
will there be "new hope"?

AlexKaravaev (approved):

PR #346, opened 2025-01-03
Introduce Opentelemetry

Michael-Preibisch:


telemetry/src/lib.rs
```diff
@@ -68,24 +195,44 @@ impl TelemetryConfig {
...
+            // This ensures that spans are flushed before exit
+            global::shutdown_tracer_provider();
+        }
     }
 }
+
+pub async fn shutdown_tracing() {
+    // Ensure all spans are flushed
+    global::shutdown_tracer_provider();
+}
```
TheButlah, line 238: configure your editor to insert final newlines and trim trailing newlines by default. See these settings in vscode:
Screenshot 2025-01-03 at 14 00 32

telemetry/src/lib.rs
```diff
@@ -46,20 +81,112 @@ impl TelemetryConfig {
...
+        // Use W3C propagation
+        global::set_text_map_propagator(TraceContextPropagator::new());
+
+        Ok(tracer_provider)
+    }
+
+    /// Try to initialize telemetry (journald/stderr + optional OTLP).
+    /// Returns an error if something goes wrong setting up the subscriber stack.
+    pub fn try_init(mut self) -> Result<(), tracing_subscriber::util::TryInitError> {
+        // 1. If OTLP was requested, set up the tracer provider
```
TheButlah, line 161: lets remove the 1,2,3 chatgpt comments

telemetry/src/lib.rs
```diff
@@ -46,20 +81,112 @@ impl TelemetryConfig {
...
+        self.service_name = service_name;
+        self.service_version = service_version;
+        self.environment = environment;
+        self
+    }
+
+    /// Initialize the OpenTelemetry TracerProvider and set it globally, returning it for storing.
+    fn init_opentelemetry(&mut self) -> Result<opentelemetry_sdk::trace::TracerProvider, Box<dyn std::error::Error>> {
+        // Fallback to environment variables if the user did not supply them.
+        let default_service_name =
```
TheButlah, line 104: all configuration for a library should be passed into the library explicitly, not through environment variables. End applications should have control over which environment variables they read and its more clear to have that in the main function of the users of this code.

telemetry/src/lib.rs
```diff
@@ -46,20 +81,112 @@ impl TelemetryConfig {
...
     }
 
-    pub fn try_init(self) -> Result<(), tracing_subscriber::util::TryInitError> {
+    /// Enable OpenTelemetry/OTLP tracing.
+    /// You can optionally provide a service name, version, and environment.
+    /// If omitted, these will default to environment variables (`SERVICE_NAME`, `SERVICE_VERSION`, `ENVIRONMENT`) or hard-coded strings.
+    #[must_use]
+    pub fn with_opentelemetry(
+        mut self,
+        service_name: Option<String>,
```
TheButlah, line 90: lets make providing a service name along with version and environment mandatory. I cant imagine not wanting that info.

telemetry/src/lib.rs
```diff
@@ -2,13 +2,39 @@ use std::io::IsTerminal as _;
...
+/// A struct controlling how telemetry will be configured (logging + optional OpenTelemetry).
 #[derive(Debug)]
 pub struct TelemetryConfig {
     syslog_identifier: Option<String>,
     global_filter: EnvFilter,
+
+    /// If true, enable OTLP tracing via OpenTelemetry.
+    use_otel: bool,
+
+    /// The service name used in opentelemetry's `Resource`.
```
TheButlah, line 29: all the open telemetry settings should live in one OpenTelemetryConfig struct, with a single `otel: Option<OpenTelemetryConfig>` in the `TelemetryConfig` struct.

telemetry/src/lib.rs
```diff
@@ -68,24 +195,44 @@ impl TelemetryConfig {
...
-    /// will cause a panic.
+    /// Initializes telemetry, panicking if something goes wrong.
     pub fn init(self) {
-        self.try_init().expect("failed to initialize orb-telemetry")
+        self.try_init().expect("failed to initialize orb-telemetry");
+    }
+
+    /// Optional shutdown hook to flush any pending OTLP spans.
+    /// For journald/stderr, it's usually not necessary.
+    pub fn shutdown_tracing(&self) {
```
TheButlah, line 227: I think the right way to do this is to instead have the `init` function return a struct called `TelemetryShutdownHandler`. On `Drop` we will call `global::shutdown_tracer_provider()`. We should then have a `#[must_use]` attribute on the type too, to ensure that people actually hang onto it.

This also allows us to remove the tracer provider from the config struct. This keeps the config struct a "builder", it isn't designed to hold state after init.

TheButlah (changes requested):

telemetry/src/lib.rs
Michael-Preibisch: I use IntelliJ RustRover mostly, but this option is there as well (probably)

telemetry/src/lib.rs
```diff
@@ -46,13 +108,79 @@ impl TelemetryConfig {
...
+        global::set_tracer_provider(tracer_provider.clone());
+        global::set_text_map_propagator(TraceContextPropagator::new());
+
+        Ok((tracer_provider, tracer))
+    }
+
+
+    /// Try to initialize telemetry (journald/stderr + optional OTLP).
+    /// Returns an error if something goes wrong setting up the subscriber stack.
+    pub fn try_init(self) -> Result<(TelemetryShutdownHandler, Result<(), tracing_subscriber::util::TryInitError>), Box<dyn std::error::Error>> {
```
TheButlah, line 161: Ths type is quite strange. I think you can simplify it by using `thiserror` for your error type, and the type signature then becomes:

```rust
pub fn try_init(self) -> Result<TelemetryShutdownHandler, OrbTelemetryErr> {
    // ...
}
```

telemetry/src/lib.rs
```diff
@@ -2,42 +2,104 @@ use std::io::IsTerminal as _;
...
+    otel: Option<OpenTelemetryConfig>,
+}
+
+/// Handles cleanup of telemetry resources on drop.
+#[must_use]
+pub struct TelemetryShutdownHandler;
+
+impl Drop for TelemetryShutdownHandler {
+    fn drop(&mut self) {
+        global::shutdown_tracer_provider();
```
TheButlah, line 61: what happens if we aren't using opentelemetry at all (for example, if we only did `with_journald` but did not do `with_opentelemetry`)? Will this panic?

update-agent/src/main.rs
```diff
@@ -49,8 +49,13 @@ const CFG_ENV_VAR: &str = const_format::concatcp!(ENV_VAR_PREFIX, "CONFIG");
 const SYSLOG_IDENTIFIER: &str = "worldcoin-update-agent";
 
 fn main() -> UpdateAgentResult {
-    orb_telemetry::TelemetryConfig::new()
+    let _telemetry_guard = orb_telemetry::TelemetryConfig::new(
+        SYSLOG_IDENTIFIER,
+        env!("CARGO_PKG_VERSION"),
```
TheButlah, line 54: use `orb-build-info` crate

TheButlah (changes requested):

telemetry/src/lib.rs
Michael-Preibisch:
```
No, it won't panic. The code is designed to safely handle cases where OpenTelemetry isn't used. If you only configure journald logging, the telemetry system will simply initialize without the OpenTelemetry components. This is achieved through Rust's Option type - when OpenTelemetry isn't configured, those components remain as None values and are safely skipped during initialization. The system will proceed with just the journald logging functionality working as expected.
This flexible design allows users to choose which telemetry features they need without worrying about dependencies on other components. You can start with basic logging and add OpenTelemetry later if needed, or just stick with logging if that's all your application requires.
```

PR #349, opened 2025-01-13
feat(update-agent): publishing update-agent progress to dbus

sfikastheo:
* `orb_update_agent_dbus` crate adds a dbus interface and proxy for sharing update-progress.
* Changes to the `orb_update_agent`crate for publishing the update progress to dbus (Best effort approach following the already archetype in the update-agent).


update-agent/dbus/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+///
+/// This trait is implemented by types that can provide information about the current update status.
+pub trait UpdateProgressT: Send + Sync + 'static {
+    fn status(&self) -> UpdateStatus;
+}
+
+/// A wrapper struct for types implementing [`UpdateProgressT`].
+pub struct UpdateProgress<T: UpdateProgressT>(pub T);
+
+#[derive(Debug, Serialize, Deserialize, Type, Clone, Value, OwnedValue)]
```
TheButlah, line 26: Might as well make it `Copy`
```suggestion
#[derive(Debug, Serialize, Deserialize, Type, Clone, Copy, Value, OwnedValue)]
```

update-agent/src/main.rs
```diff
@@ -252,6 +278,15 @@ fn run(args: &Args) -> eyre::Result<()> {
         info!("running update for component `{}`", component.name());
         component
             .run_update(target_slot, &claim, settings.recovery)
+            .inspect(|_| {
+                if let Some(conn) = dbus_update_conn {
+                    interfaces::update_component_state(
+                        component.name(),
+                        ComponentState::Installed,
+                        conn,
```
TheButlah, line 286: try passing around the InterfaceRef instead of the connection itself. It makes it a bit more clear what this function is doing.

TheButlah (changes requested):
very nice! just a few small tweaks and we will be done :)

update-agent/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,81 @@
...
+pub fn signal_error(error: &str, conn: &Connection) {
+    if let Ok(iface) = get_iface_ref(conn) {
+        iface.get_mut().0.error = error.to_string();
+        emit_signal(&iface);
+    } else {
+        warn!("failed to get interface reference");
+    }
+}
+
+pub fn init_components(components: &[ManifestComponent], conn: &Connection) {
```
TheButlah, line 36: the name could be more informative, maybe `initialize_dbus_properties`

update-agent/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,81 @@
...
+                name: c.name.clone(),
+                state: ComponentState::None,
+            })
+            .collect();
+    } else {
+        warn!("failed to get interface reference");
+    }
+}
+
+pub fn update_component_state(name: &str, state: ComponentState, conn: &Connection) {
```
TheButlah, line 50: name could be more informative, maybe `set_dbus_properties`

update-agent/src/main.rs
```diff
@@ -252,6 +278,15 @@ fn run(args: &Args) -> eyre::Result<()> {
         info!("running update for component `{}`", component.name());
         component
             .run_update(target_slot, &claim, settings.recovery)
+            .inspect(|_| {
```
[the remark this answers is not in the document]
sfikastheo: Yes, announcing the progress must be best effort, not bring the update-agent down

update-agent/dbus/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+pub enum ComponentState {
+    None = 1,
+    Fetched = 2,
+    Processed = 3,
```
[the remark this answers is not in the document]
sfikastheo: No, you are fetching the compressed version, you process it ( check hash, decompress, check hash ).  Just used the already terminology that the update-agent is using

update-agent/dbus/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// Represents the overall update status.
+///
+/// Provides detailed state of the update progress for each component.
+/// In case of failure, the error message is stored in the `error` field.
+#[derive(Default, Debug, Serialize, Deserialize, Type, Clone, Value, OwnedValue)]
+pub struct UpdateStatus {
+    pub components: Vec<ComponentStatus>,
+    pub error: String,
```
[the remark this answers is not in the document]
sfikastheo: As discussed the idea here is that you share the state of the update thus far. If you use result, and you fail, the `UpdateProgress` will only contain the error message without the state of the update at the time of the error.
Can be swayed either way, I agree it would be more "beautiful" rust.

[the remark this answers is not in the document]
TheButlah: i think this tradeoff is acceptable to make it more idiomatic, and make it more likely that the calling service realizes that there is an error and is forced to check that case.

If you think the progress reporting at time of error is extremely important, these messages will still show up in logs and additionally you could include it in the string message of the error, if you so choose.

update-agent/dbus/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+//! dbus-monitor --session type='signal',sender='org.worldcoin.UpdateAgentManager1'
+//! ```
+
+use serde::{Deserialize, Serialize};
+use zbus::interface;
+use zbus::zvariant::{OwnedValue, Type, Value};
+
+/// A trait representing update progress behavior.
+///
+/// This trait is implemented by types that can provide information about the current update status.
```
TheButlah, line 20: maybe explain more clearly why the trait and wrapper type exist. Its to allow dependency injection, mocking, and sharing the same types across both client and server.

update-agent/dbus/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    Installed = 5,
+}
+
+#[derive(
+    Debug, Serialize, Deserialize, Type, Eq, PartialEq, Clone, Value, OwnedValue,
+)]
+pub struct ComponentStatus {
+    pub name: String,
+    pub state: ComponentState,
+    pub progress: u8,
```
TheButlah, line 45: add comment - is it 0-100?

update-agent/src/component.rs
```diff
@@ -292,13 +301,15 @@ fn extract<P: AsRef<Path>>(path: P, uncompressed_download_path: P) -> eyre::Resu
 }
 
 #[expect(clippy::result_large_err)]
+#[allow(clippy::too_many_arguments)]
```
TheButlah, line 304: only ever use `expect`

update-agent/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,58 @@
...
+        .0
+        .components
+        .iter_mut()
+        .find(|c| c.name == name)
+        .map(|component| {
+            component.state = state;
+            component.progress = progress;
+        })
+        .unwrap_or_else(|| {
+            warn!("Component {} not found", name);
```
TheButlah, line 50: probably if the updating fails, we should return an err to the caller. Not log. let the caller decide if they should log.

update-agent/src/dbus/interfaces.rs
```diff
@@ -0,0 +1,58 @@
...
+        });
+
+    if let Err(e) =
+        zbus::block_on(iface.get_mut().progress_changed(iface.signal_context()))
+    {
+        warn!("Failed to emit progress_changed signal: {}", e);
```
TheButlah, line 56: same here, let the caller decide if they should log or how to handle the error

update-agent/src/main.rs
```diff
@@ -77,6 +84,55 @@ fn get_config_source(args: &Args) -> Cow<'_, Path> {
...
+
+    let dbus_conn = match setup_conn() {
+        Ok(conn) => conn,
+        Err(e) => {
+            warn!("failed to setup dbus connection: {e:?}");
+            return (None, None);
+        }
+    };
+
+    // Build a dbus proxy for the suprevisor service
```
TheButlah, line 113:
```suggestion
    // Build a dbus proxy for the supervisor service
```

update-agent/src/main.rs
```diff
@@ -77,6 +84,55 @@ fn get_config_source(args: &Args) -> Cow<'_, Path> {
...
+    let supervisor_proxy = proxies::SupervisorProxyBlocking::builder(&dbus_conn)
+        .cache_properties(zbus::CacheProperties::No)
```
TheButlah, line 115: explain why in a comment

TheButlah (changes requested):
very close to done. just a few small tweaks and we are good to go. Thanks!

update-agent/src/dbus/interfaces.rs
sfikastheo: Probably not even needed tbh

update-agent/src/dbus/interfaces.rs
sfikastheo: Fair, adds a bit of boilercode into main, but i agree. Thanks for the help :)

TheButlah (approved):

PR #350, opened 2025-01-17
feat: implement cbor size experiment

TheButlah:


TheButlah:
Closing, we have landed cbor support.

PR #351, opened 2025-01-18
fix(update-agent): fix destination on xz components (ORBS-241)

sfikastheo:
Fix on [PR](https://github.com/worldcoin/orb-software/pull/330):
The dst that was passed into `component.process()`, made the `xz` components always look for the `dst` instead the component specific file under `dst/component-unique-name`.
This was forcing all the compressed components to `reprocess` and actually due to the size of the `downloads` directory the `size` check in the CAN components was constantly failing the update-agent.

Additionally we were not following the notion of `.verified` file flags for the `uncompressed` components but were overwriting the one that already existed from the component fetching process.

### DIFF
--- a/update-agent/src/component.rs
+++ b/update-agent/src/component.rs
@@ -83,9 +83,11 @@ impl Component {
     }
 
     fn process_compressed(&mut self, dst: &Path) -> eyre::Result<()> {
-        let uncompressed_path = dst.with_extension("uncompressed");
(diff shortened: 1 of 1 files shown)


AlexKaravaev:
I am not that deep into update-agent code, but change seems to be ok. If tested - then good to go

AlexKaravaev (approved):

alekseifedotov (approved):

PR #354, opened 2025-01-22
feat(telemetry): add opentelemetry support in orb-telemetry

TheButlah:
Adds tracing-opentelemetry as a tracing subscriber and also uses opentelemeyry-otlp to export spans to an opentelemetry collector

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4903,6 +4903,71 @@ dependencies = [
 ]
 
 [[package]]
+name = "opentelemetry"
+version = "0.27.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


TheButlah:
I actually based it on https://github.com/worldcoin/orb-software/commit/cb9975e567bf3c5b4674730ff13f470d8130d078

TheButlah:
Great! Tomorrow I have a meeting with @valff to make sure this works how he wants it, then we can merge (and likely also use it in orb-core).

TheButlah:
Need @valff to review so that we can start using this in orb-core too.

TheButlah:
@valff is pretty swamped next week, so I've gotten his permission to merge into orb-core *if* we get reviews from both @karelispanagiotis and @bsayed

karelispanagiotis (approved):

bsayed (approved):
If we have decided to move away from Datadog then LGTM

PR #356, opened 2025-01-29
build: add docker runner for mac users to run tests

TheButlah:
lets mac users run tests that otherwise only run on linux.

### DIFF
--- a/docker/cargo-runner.sh
+++ b/docker/cargo-runner.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Allows running tests inside docker. Takes care of linux-only dependencies and avoids
+# the need to emulate the architecture.
+
+set -o pipefail -eu
+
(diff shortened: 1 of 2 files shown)


docker/cargo-runner.sh
```diff
@@ -0,0 +1,21 @@
...
+# the need to emulate the architecture.
+
+set -o pipefail -eu
+
+PROGRAM=$1; shift
+
+main() {
+	local -r absolute_path="$(realpath ${PROGRAM})"
+
+	docker run \
```
AlexKaravaev, line 12: do we want to reuse test-utils crate for this?

AlexKaravaev (approved):

docker/cargo-runner.sh
TheButlah: probably not imo

PR #358, opened 2025-01-30
feat: wire up cellcom to backend

oldgalileo:


Qbicz:
Cool! Where you able to get some location based on serving cell info?
When I quickly tried with OpenCellId it failed so far (cell not found)

cellcom/Cargo.toml
```diff
@@ -8,6 +8,11 @@ license.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 
+# TODO: Switch fully between reqwest for Google and orb-security-utils for Status backends
+[features]
```
TheButlah, line 12: blocker: we should not use cargo feature flags for customizing backend endpoints. Use `ORB_BACKEND` env var and switch at runtime. or better yet, explicitly accept as an argument the backend type.

cellcom/Cargo.toml
```diff
@@ -18,3 +23,5 @@ serialport = "4.6.1"
 tracing.workspace = true
 tracing-subscriber.workspace = true
 wpactrl = "0.5.1"
+orb-endpoints.workspace = true
+orb-security-utils = { workspace = true, features = [ "reqwest", "blocking" ] }
```
TheButlah, line 27: I think all code we write should generally use async for stuff that does io. But I don't want to drag down the PR. So I will not block it.

TheButlah (changes requested):

cellcom/src/backend/status.rs
```diff
@@ -0,0 +1,244 @@
...
+    wifi_networks: &[WifiNetwork],
+) -> Result<()> {
+    let request = build_status_request(orb_id, cellular_info, wifi_networks)?;
+
+    debug!(
+        request = ?serde_json::to_string(&request)?,
+        "Sending status v2 request"
+    );
+
+    let backend = Backend::from_env_or_build_type::<{ cfg!(feature = "stage") }>();
```
TheButlah, line 66: I am gonna remove this function, it was a mistake. Doesn't make sense anymore after the analysis backend was added. Can you use `ORB_BACKEND` or explicitly accept the backend enum as an arg?

cellcom/src/backend/status.rs
```diff
@@ -0,0 +1,244 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct CellDataV2 {
+    pub mcc: Option<u32>,
+    pub mnc: Option<u32>,
+    pub lac: Option<String>,
+    pub cell_id: Option<u32>,
+}
+
+pub fn client() -> &'static Client {
```
TheButlah, line 44: @vmenge has been pushing against using this approach, especially in libraries. IMO the client should not be constructed in this helper library, it should be passed in via the configuration for the library (along with the backend type).

This will make for more testable code.

cellcom/src/backend/status.rs
```diff
@@ -0,0 +1,259 @@
...
+    debug!(
+        request = ?serde_json::to_string(&request)?,
+        "Sending status v2 request"
+    );
+
+    let backend = Backend::from_env()?;
+    // TODO: Some sort of actual response besides the status code
+    // maybe one that includes the location? (requires changes on the
+    // backend)
+    let _response_raw = client()
```
TheButlah, line 84: you should still check status code

TheButlah (changes requested):

TheButlah (approved):

PR #364, opened 2025-02-03
fix(upload-certs): specify active=true

sfikastheo:


scripts/upload-certs.sh
```diff
@@ -124,7 +124,8 @@ main() {
             "type": "signup",
             "key": "'"${signup_pubkey}"'",
             "signature": "'$(base64 -w 0 -i "${keypath}/70000002.signature.raw")'",
-            "extraData": "'$(base64 -w 0 -i "${keypath}/70000002.extra.raw")'"
+            "extraData": "'$(base64 -w 0 -i "${keypath}/70000002.extra.raw")'",
```
oldgalileo, line 127: You're encoding. No need for the `-i

oldgalileo (changes requested):

scripts/upload-certs.sh
sfikastheo: Nice catch thanks. Going to remove it from everything

oldgalileo (approved):

PR #366, opened 2025-02-06
feat(orb-info): FM-370-orb-info

paulquinn00:
Crate of helpers to retrieve common Orb info values.
Currently supported:
- orb id
- orb name
- orb jabil id
- orb auth token

Additionally:
- Moved `orb_endpoints::OrbId` to `orb_info::OrbId`


orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,155 @@
...
+impl OrbToken {
+    pub async fn new() -> Result<Self, OrbInfoError> {
+        let (join_handle, auth_token_receiver) = Self::setup_dbus().await?;
+        Ok(OrbToken {
+            auth_token_receiver,
+            join_handle,
+        })
+    }
+
+    pub async fn get_orb_token(&self) -> Result<String, OrbInfoError> {
```
TheButlah, line 38: instead of a struct which provides an api for accessing tokens, consider instead just providing the necessary functionality to spawn the tokio task that listens for a token, and explicitly expose to the caller a tokio::sync::watch::Receiver<OrbToken>, plus any additional tools the caller may need to manage the lifecycle of the task.

Example:
```rust
pub struct TokenWatchTask {
    pub join_handle: tokio::task::JoinHandle<Result<()>>,
    pub token_watcher: tokio::sync::watch::Receiver<OrbToken>,
}

impl TokenWatchTask {
    pub fn spawn(cancel_token: tokio_util::sync::CancelationToken) -> (join, watcher) {
        let (kill_tx, kill_rx) = tokio::sync::oneshot::channel();
        let join_handle = tokio::task::spawn(Self::task_inner(kill_rx));
    }

    async fn task_inner(kill_rx: tokio::sync::oneshot::Receiver<()>) -> Result<()> {
        let do_stuff_fut = async {
            loop { 
                ///

            }
        };

        tokio::select! {
            _ = cancel_token.cancelled() = Ok(()),
            result = do_stuff_fut => result,

        }

    }
}

```

TheButlah (changes requested):

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,165 @@
...
+        let join_handle =
+            tokio::task::spawn(Self::task_inner(auth_token_proxy, send, cancel_token));
+        Ok((join_handle, recv))
+    }
+
+    async fn task_inner(
+        auth_token_proxy: AuthTokenManagerProxy<'_>,
+        send: watch::Sender<String>,
+        cancel_token: tokio_util::sync::CancellationToken,
+    ) {
```
TheButlah, line 57: Should return `color_eyre::Result<()>` so that we can allow the errors to get propagated via the `JoinHandle<Result<()>>`

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,165 @@
...
+        let token_updater = async move {
+            let mut token_changed = auth_token_proxy.receive_token_changed().await;
+            while let Some(token) = token_changed.next().await {
+                let token = token.get().await.expect("should have received token");
```
TheButlah, line 61:
```suggestion
                let token = token.get().await.wrap_err("failed to get token over dbus")?;
```

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,165 @@
...
+                send.send(token)
+                    .expect("should have sent token to watchers");
```
TheButlah, line 63: Its normal for this to fail if for example, all watchers are dropped. Instead of panicking, you should *cleanly* `return Ok(());`, allowing the task to exit cleanly / normally.

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+        };
+
+        tokio::select! {
+            _ = cancel_token.cancelled() => {
+                println!("cancelled");
```
TheButlah, line 69:
```suggestion
                debug!("Cancelling token watcher task");
```

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+            _ = token_updater => {
+                println!("token updater");
+            }
```
TheButlah, line 73:
```suggestion
            result = token_updater => result,
```
return the async future's evaluated value as the result of the function. This lets you return things like Err(..) to return an error from the function, etc

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,165 @@
...
+    use std::{sync::Arc, time::Duration};
+    use tokio::sync::Mutex;
+    use zbus::{object_server::SignalContext, ConnectionBuilder};
+
+    #[derive(Clone)]
+    struct TestAuthTokenManager {
+        token: Arc<Mutex<String>>,
+    }
+
+    #[zbus::interface(name = "org.worldcoin.AuthTokenManager1")]
```
TheButlah, line 90: instead of re-creating a zbus interface, use the existing `orb_attest_dbus::AuthTokenManager` struct. This struct exists to provide a "wrapper" type that implements all the zbus macros, and allows you to use dependency injection to instantiate it with anything that implements `orb_attest_dbus::AuthTokenManagerT`.

I apologize for not having sufficient code examples for this already in the repo - its a relatively new pattern we are using. A lot of our dbus code simply doesn't have tests, so I'm very glad that you are actually testing the code! Thank you :)

For now, here is some example code to showcase what I mean, see if you can incorporate these suggestions into your PR.

```rust
#[derive(Clone, Debug)]
struct Mocked {
    token: Arc<Mutex<String>>,
}

// Note how we are simply implementing a trait from orb-attest-dbus instead of creating an entirely new struct with zbus macros.
// This ensures that the function signatures all match up and we get good compile errors and LSP support.
impl orb_attest_dbus::AuthTokenManagerT for Mocked {
        async fn token(&self) -> String {
            self.token.lock().await.clone()
        }
}

// using `dbus_launch` ensures that all tests use their own isolated dbus, and that they can't influence each other.
async fn start_dbus_daemon() -> dbus_launch::Daemon {
    tokio::task::spawn_blocking(|| dbus_launch::Launcher::daemon()
        .launch()
        .expect("failed to launch dbus-daemon")
    )
    .await
    .expect("task panicked")
}


async fn setup_test_server() -> Result<(Connection, dbus_launch::Daemon, Mocked)> {
        let token = Arc::new(Mutex::new("test_token".to_string()));
        let mock_manager = Mocked { token };
        let dbus_daemon = start_dbus_daemon().await;

        // instead of using the "Default" dbus, we explicitly use the one from the daemon we spawned.
        // its also important that we can then pass this zbus::Connection along to where we instantiate the
        // AuthTokenProxy in our tests, so that they are talking on the same bus.
        // Thats why I recommended that you explicitly pass the connection into the `spawn/read` function on line 19.
        let connection = ConnectionBuilder::address(dbus_daemon.address())?
            .name("org.worldcoin.AuthTokenManager1")?
            // see how the dependency injection happens here?
            .serve_at("/org/worldcoin/AuthTokenManager1", orb_attest_dbus::AuthTokenManager(mock_manager))?
            .build()
            .await?;

        Ok((connection, dbus_daemon, mock_manager))
}

```

.github/workflows/rust-ci.yaml
```diff
@@ -128,7 +128,9 @@ jobs:
           key: custom-${{ hashFiles('**/*.nix', 'flake.lock') }}
       - name: Install dbus
         if: matrix.platform == 'macos-14'
-        run: brew install dbus
+        run: |
+          brew install dbus
+          brew services start dbus
```
TheButlah, line 133: Question: Do we still need this if use https://docs.rs/dbus_launch?

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,165 @@
...
+            .name("org.worldcoin.AuthTokenManager1")?
+            .serve_at("/org/worldcoin/AuthTokenManager1", test_manager.clone())?
+            .build()
+            .await?;
+
+        Ok((connection, test_manager))
+    }
+
+    #[tokio::test]
+    #[serial_test::serial]
```
TheButlah, line 112: probably doesn't need `serial_test` if we use `dbus_launch`.

TheButlah (changes requested):

.github/workflows/rust-ci.yaml
paulquinn00: nope, removed

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct Token(String);
+
+#[derive(Debug)]
+pub struct OrbToken {
+    auth_token_receiver: watch::Receiver<Token>,
+    join_handle: tokio::task::JoinHandle<()>,
+}
+
+impl OrbToken {
+    pub async fn new() -> Result<Self, OrbInfoError> {
```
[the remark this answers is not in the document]
paulquinn00: What if the client sends in the wrong dbus connection? This seems fraught with issues.

.github/workflows/rust-ci.yaml
paulquinn00: nm, ci still fails without this

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,203 @@
...
+        .await
+        .expect("task panicked")
+    }
+
+    async fn setup_test_server() -> Result<(Connection, dbus_launch::Daemon, Mocked)> {
+        let token = Arc::new(Mutex::new("test_token".to_string()));
+        let mock_manager = Mocked { token };
+        let daemon = start_dbus_daemon().await;
+
+        let connection = ConnectionBuilder::session()?
```
TheButlah, line 130:
```suggestion
        let connection = ConnectionBuilder::address(daemon.address())?
```
without this, dbus uses the default system dbus. Thats probably why the tests were requiring serial as well as the `brew` dbus stuff.

Remove any serial attributes on tests unless it is intending to mutate environment variables (which orb-token should not).

Also, don't forget to stop dropping the daemon (don't assign to `_`) because that will kill the daemon.

TheButlah (changes requested):

.github/workflows/rust-ci.yaml
TheButlah: I'm pretty sure that this is unecessary, try using dbus-launch's address in the zbus connection builder.

orb-info/src/orb_token.rs
paulquinn00: I did not know that `_var` would keep it alive. Thought it was scoped by usage. Nice!

orb-info/src/orb_token.rs
```diff
@@ -0,0 +1,200 @@
...
+            .token_changed(iface_ref.signal_context())
+            .await
+            .wrap_err("failed to send token_changed signal")?;
+
+        // Verify updated token is retrieved
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        let updated_token = orb_token.value();
+        assert_eq!(updated_token, "updated_token");
+
+        // connection.close().await?;
```
TheButlah, line 197:
```suggestion
```

TheButlah (approved):

TheButlah:
Awesome work @paulquinn00 ! Thanks for the patience during the review process. Approved.

PR #368, opened 2025-02-07
feat(mcu-update): increase max mcu firmware size

fouge:
diamond's main mcu uses an external flash that allows to use a larger image size, up to 448KiB.

### DIFF
--- a/update-agent/src/update/can.rs
+++ b/update-agent/src/update/can.rs
@@ -79,9 +79,9 @@ const UPDATE_AGENT_ISOTP_ID: IsoTpNodeIdentifier = IsoTpNodeIdentifier::JetsonAp
 /// This constant is used to register interest in an event on the event poller. The API makes this
 /// necessary but the value carries no extra meaning.
 const ARBITRARY_EVENT_KEY: usize = 42;
-/// MCU_MAX_FW_LEN_BYTES is 224KiB (per slot), the absolute maximum length
-/// an MCU update can be. This is defined by the [MCU board DTS](https://github.com/worldcoin/orb-mcu-firmware/blob/d98719185b59375429123a5fd275dd5696a5bf12/boards/arm/mcu_main/mcu_main.dts#L516)
-const MCU_MAX_FW_LEN_BYTES: u64 = 224 * 1024;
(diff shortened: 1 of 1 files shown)


update-agent/src/update/can.rs
```diff
@@ -79,9 +79,9 @@ const UPDATE_AGENT_ISOTP_ID: IsoTpNodeIdentifier = IsoTpNodeIdentifier::JetsonAp
 /// This constant is used to register interest in an event on the event poller. The API makes this
 /// necessary but the value carries no extra meaning.
 const ARBITRARY_EVENT_KEY: usize = 42;
-/// MCU_MAX_FW_LEN_BYTES is 224KiB (per slot), the absolute maximum length
-/// an MCU update can be. This is defined by the [MCU board DTS](https://github.com/worldcoin/orb-mcu-firmware/blob/d98719185b59375429123a5fd275dd5696a5bf12/boards/arm/mcu_main/mcu_main.dts#L516)
-const MCU_MAX_FW_LEN_BYTES: u64 = 224 * 1024;
+/// MCU_MAX_FW_LEN_BYTES is 448KiB (per slot), the absolute maximum length
+/// an MCU update can be. This is defined by the MCUs' partitioning scheme.
+const MCU_MAX_FW_LEN_BYTES: u64 = 448 * 1024;
```
sfikastheo, line 84: The same check is also used in `pearl`, to my understanding the external SPI is a diamond only feature.
Are we sure that we want to invalidate the check for `pearl` ?
Would we opt for a runtime check of the hardware-revision to set the correct threshold if it was an option?

(Although i am not a huge fan of adding more fragile external state into the the `update-agent`)

sfikastheo: approving, and leaving it up to you Cyril :bow:

sfikastheo (approved):

update-agent/src/update/can.rs
fouge: yeah it's fine. This check will prevent sending any other component by accident (very likely larger than 448KiB). It only ensure we are actually sending a firmware image (will being a very loose check)

The size is also checked on the MCU side and returns an error when sending the first block if it cannot fit the receiving slot.

PR #370, opened 2025-02-09
feat(fleet-cmdr): FM-369 using orb-info for orb details values

paulquinn00:
Updates to fleet-cmdr:
- handle JobNotify and JobRequestNext
- utilize `orb-info` crate for common values:
  - orb id, orb name, jabil id, orb token


TheButlah (commented):
Unfortunately I don't have time rn to give a real review, sorry @paulquinn00 but it will probably have to wait until after diamond 😢

PR #374, opened 2025-02-13
feat(nix): remove nixos-generators, add disko

TheButlah:
This PR:
* adds support for [disko](https://github.com/nix-community/disko) which handles declarative partitioning of our HILs and Liveusb. We no longer need to struggle with the manual partitioning process when installing the OS.
* Uses disko instead of nixos-generators, to lay out the partitions in a way that actually will work on GPT-and-UEFI-only machines (like the asus NUC and lattepanda).
* Implements CI automation for producing the liveusbs via disko, so that anyone can just download the liveusb
* Fixes the liveusb to actually work
* Upstreams everything we need from https://github.com/thebutlah/scratch
* removes nix-channels, instead pinning against the flake
* sets up getty, so that the HILs don't require a password prompt for the tty. Makes it possible to use them without needing SSH, because mutableUsers = false and no other password settings implies there is no password to type at the tty.
* CI also builds all the hils (although that was mostly already set up)

Future work needed:
* figure out how to get the liveusb to be immutable/a squashfs.
* Update the github runner


nix/machines/README.md
```diff
@@ -8,3 +8,8 @@ under `/etc/nixos/` on a NixOS machine.
 
 These are then used by `flake-outputs.nix` and combined with the toplevel
 `flake.nix`.
+
+## Building liveusb
+```bash
+
```
AlexKaravaev, line 14: empty :(

PR #383, opened 2025-02-21
feat(wt_video): implement demo

TheButlah:
This demos using webtransport to perform video streaming from the "orb" to the browser. It also demonstrates some ideas I had for efficient pools of frame buffers, as well as experiments with png encoding.

I think ultimately instead of png encoding, we should use an actual video codec + use the browser [WebCodecs](https://developer.mozilla.org/en-US/docs/Web/API/WebCodecs_API) api, instead of doing the jank png encoding. But even so, this works!


AlexKaravaev (approved):
very nit, but maybe README.md for every demo to not forget about it can be helpful, but LGTM

PR #389, opened 2025-02-25
feat: add backend shared crate to orb-attest

AlexKaravaev:
just was doing something else and noticed small cleanup chance


TheButlah (approved):
nice! don't forget to get CI to pass. You need to `git add` your lockfile.

PR #393, opened 2025-02-27
feat(gst-jpeg): hardware-accelerated jpg encode

TheButlah:
Used in orb-core livestream. In the future, it would be nice to detect the supported plugins and chose the ones supported on the host platform, instead of just being nvidia-only.

### DIFF
--- a/gst-jpeg/src/lib.rs
+++ b/gst-jpeg/src/lib.rs
@@ -1,0 +1,99 @@
+use eyre::{bail, Result};
+use gstreamer::{
+    prelude::{Cast as _, ElementExt as _, GstBinExtManual as _},
+    ElementFactory, Pipeline,
+};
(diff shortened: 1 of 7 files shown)


gst-jpeg/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+                    frame.len(),
+                );
+            }
+        }
+        self.appsrc.push_buffer(buffer)?;
+        // NOTE(@thebutlah): we could have used the callback API but I think this way
+        // is simpler, since it keeps everything synchronous and the api for the caller
+        // is simple.
+        let sample = self.appsink.pull_sample()?;
+        trace!(?sample, "got sample");
```
AlexKaravaev, line 86: not sure if it will spam too much, maybe we can throttle

AlexKaravaev (approved):

PR #401, opened 2025-03-07
feat(ui): white during idle / user scanning (self-serve) 

fouge:
TODO

- [x] Thomas comment: Let’s go down in brightness 25%. 
    - @fouge : changed from 40 to 30. **But a problem arises, see below.**
- [ ] Thomas comment: Can we fade in as slow as it fades out?
   - @fouge's comment: same time right now between in and out, but it's not noticeable: on pearl, the intensity is hard to control without bad UX: we notice the intensity **steps** when pulsing, on values close to 0. The animation should be smoother. More details: the current value goes from 1 to 30 using a sine wave. The intensity 29 to 30 is hardly noticeable if noticeable at all by the human eye (looks solid white), while the intensity from 1 to 2 is very noticeable (too much).
    - **Let's use half the LEDs** to decrease general brightness: One idea to make it smooth: switch between which LEDs are on for each new frame. this might help to make sure we don't notice that we use only half the LED, because the light diffusor on Pearl isn't that strong [I would reset half of the LEDs to intensity 0 right before we send the values to the MCU.](https://github.com/worldcoin/orb-software/blob/d50fabfc4419e667f192ff1cdd22fe8f5e41f558/ui/src/engine/pearl/mod.rs#L58) and leave the computation the same. 

March 7th:
@karelispanagiotis now taking over.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -8,6 +8,7 @@ ignore = [
   { id = "RUSTSEC-2023-0071", reason = "No patch available yet, we also dont plan to use rsa keys" },
   { id = "RUSTSEC-2024-0429", reason = "We need to update gstreamer in orb-core first"  },
   { id = "RUSTSEC-2024-0436", reason = "paste - no longer maintained"  },
+  { id = "RUSTSEC-2025-0014", reason = "humantime is used cross many crates" }
 ]
 
 [sources]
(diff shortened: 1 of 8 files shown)


fouge:
@karelispanagiotis i just rebased main, can you review and make sure it behaves correctly? 🤔

karelispanagiotis:
UX changes introduced in this PR:
* Operator QR scan phase & User QR scan phase are identical 
    * breathing white.
    * the only difference is in the operator LEDs.
* biometric capture progress slider is white.
* removed user-qr code backend validation sound.
* biometric capture success blinks the ring twice and then stays blank until the singup is complete.
    * it is configured from orb-core to use by default green-blinking, otherwise the blinking is white.

PR #402, opened 2025-03-08
feat(nix): build container for cargo tests

TheButlah:
I don't have a linux machine on me right now so I rely on running cargo tests. We have an existing cargo runner script that runs the tests in a debian docker container. But this is missing most of our actual dependencies. We need instead to actually build a docker image with all the necessary dependencies and use that.
https://linear.app/worldcoin/issue/ORBS-390/get-tests-to-run-in-cargo-runner-via-docker


TheButlah:
forced a merge due to lack of review, as it blocks update-agent binary diffing work.

PR #399, opened 2025-03-06
feat(ui): change qr-scan animation & color

karelispanagiotis:

PR #403, opened 2025-03-10
revert(ui): Revert "revert(ui): change qr-scan animation & color #399 (#400)"

andronat:
This reverts commit d50fabfc4419e667f192ff1cdd22fe8f5e41f558.

### DIFF
--- a/ui/rgb/src/lib.rs
+++ b/ui/rgb/src/lib.rs
@@ -101,6 +101,7 @@ impl Argb {
     /// Outer-ring color during user QR scans
     pub const PEARL_RING_USER_QR_SCAN: Argb = Argb(None, 30, 20, 0);
     pub const PEARL_RING_USER_QR_SCAN_SPINNER: Argb = Argb(None, 28, 25, 10);
+    pub const PEARL_RING_USER_QR_SCAN_WAVE: Argb = Argb(None, 255, 255, 255);
     /// Shroud color to invite user to scan / reposition in front of the orb
     pub const PEARL_CENTER_SUMMON_USER_AMBER: Argb = Argb(None, 30, 20, 0);
     /// Shroud color during user scan/capture (in progress)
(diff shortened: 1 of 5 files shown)


ui/rgb/src/lib.rs
```diff
@@ -101,6 +101,7 @@ impl Argb {
     /// Outer-ring color during user QR scans
     pub const PEARL_RING_USER_QR_SCAN: Argb = Argb(None, 30, 20, 0);
     pub const PEARL_RING_USER_QR_SCAN_SPINNER: Argb = Argb(None, 28, 25, 10);
+    pub const PEARL_RING_USER_QR_SCAN_WAVE: Argb = Argb(None, 255, 255, 255);
```
fouge, line 104: this is way too bright. Thomas M asked for 25% reduction based on value of 40. So I changed to 30 in https://github.com/worldcoin/orb-software/pull/401

ui/src/engine/pearl/self_serve.rs
```diff
@@ -148,6 +148,18 @@ impl Runner<PEARL_RING_LED_COUNT, PEARL_CENTER_LED_COUNT> {
...
+                        self.operator_signup_phase.user_qr_code_ok();
+                        self.set_ring(
+                            LEVEL_FOREGROUND,
+                            animations::Wave::new(
+                                Argb::PEARL_RING_USER_QR_SCAN_WAVE,
+                                5.0,
+                                0.5,
+                                true,
+                            ),
+                        );
```
fouge, line 161: we would need a smooth transition here, see https://github.com/worldcoin/orb-software/pull/401

fouge (changes requested):
i don't think we will ever try to use `User` or `UserV2` interchangeably (based on a backend config for example)

to me it's adding complexity and is technical debt that we know we can already clean.

PR #406, opened 2025-03-10
feat(mcu-util): control gimbal relative to current position

fouge:
allows us to go past the available range for testing gimbal


TheButlah (approved):
Should we print out when the range requested exceeds what is possible? And will this damage the hardware?

PR #414, opened 2025-03-13
feat(bidiff-squashfs): add diff command to cli

TheButlah:


bidiff-squashfs/cli/src/main.rs
```diff
@@ -63,8 +65,26 @@ fn main() -> Result<()> {
...
+        .new_path(&args.top)
+        .new(&top_contents)
+        .out(&mut out_writer)
+        .diff_params(&DiffParams::default())
+        .call()
+        .wrap_err("failed to perform diff")?;
+    out_writer
+        .into_inner()
+        .wrap_err("failed to flush buffered writer")?
+        .flush()
```
alekseifedotov, line 86: why is it need to be flushed? It should do that automatically when destroyed, shouldn't it?

alekseifedotov (approved):

bidiff-squashfs/cli/src/main.rs
TheButlah: yes, files are synced when they go out of scope. But that won't propagate errors.

TheButlah: Update: `.flush()` on a `std::fs::File` is a no-op, because they are not buffered. I should have used `sync_all()`. 
Screenshot 2025-03-13 at 16 11 38

PR #419, opened 2025-03-17
feat(location): FM-403-orbsoft-location-pkg

paulquinn00:
- Cargo toml deb target 
- debian service files


location/debian/fleet-cmdr.worldcoin-fleet-cmdr.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Worldcoin Orb Location info gatherer
+
+[Service]
+Type=simple
+ExecStart=/usr/local/bin/orb-location
```
Michael-Preibisch, line 6: Nice 🚀 We'll need to pass args to `orb-location`, I will add it to this PR

PR #423, opened 2025-03-19
Revert "feat: Orb-Location (wifi + cell location daemon binary)"

TheButlah:
Reverts worldcoin/orb-software#416


TheButlah:
hmm, I was hoping I could fix the authorship with a normal revert, but its not sufficient to fix it 😢

PR #424, opened 2025-03-19
refactor(cellcom): break into multiple crates

TheButlah:
Right now, google API code, quectl hardware wrapper, and wpa supplicant wrappers all exist in the same crate. This PR breaks out these domain-specific and independent apis into their own crates, and introduces `orb-location` to serve as the main application/location for business logic. This will empower the fleet management team to own the business logic and daemon, while keeping some of the device specifics (in particular wpa supplicant and the cellcom stuff) separately maintained.


TheButlah:
Merging so that the fm team has this stuff ready by beginning of day tomorrow, with (hopefully) minimal disruption

PR #434, opened 2025-03-20
feat(bidiff-squashfs): Implement new command to diff entire OTAs

TheButlah:
rebased on #436 

The process of crafting a bidiff ota has proven to be annoying and manual, and is bottle-necking my testing. This new command will automate the process.

In the worst case, we can use this to produce OTAs by hand, instead of in CI. Even if long term we use CI, producing them locally seems useful regardless, for testing and development.


TheButlah:
Closed and deleted in favor of incrementally merging PRs as I build it.

PR #435, opened 2025-03-21
feat(s3-helpers): implement list_prefix

TheButlah:
Introduces the orb-s3-helpers crate to consolidate some common s3 related code between orb-hil and orb-bidiff-squashfs-cli.

Also introduces localstack for testing operations against s3, and implements list_prefix which will be used as part of orb-bidiff-squashfs-cli to download directories of s3 objects.

### DIFF
--- a/s3-helpers/src/client.rs
+++ b/s3-helpers/src/client.rs
@@ -1,0 +1,65 @@
+use aws_config::{
+    meta::{credentials::CredentialsProviderChain, region::RegionProviderChain},
+    retry::RetryConfig,
+    stalled_stream_protection::StalledStreamProtectionConfig,
+    BehaviorVersion,
+};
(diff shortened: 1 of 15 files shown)


s3-helpers/src/client.rs
```diff
@@ -0,0 +1,65 @@
...
+use futures::TryStream;
+use tracing::info;
+
+use crate::s3_url_parts::S3Uri;
+
+const TIMEOUT_RETRY_ATTEMPTS: u32 = 5;
+
+/// Helper function for setting up aws credentials.
+pub async fn client() -> Result<aws_sdk_s3::Client> {
+    let region_provider = RegionProviderChain::default_provider().or_else("us-east-1");
```
AlexKaravaev, line 18: nit: we might want to put "us-east-1" into constants above?

AlexKaravaev (approved):

PR #442, opened 2025-03-26
feat(bidiff-squashfs): implement diffing plan, clean up fetching

TheButlah:
In order to diff the OTA, we need to go through all the files and generate a plan of action. This PR introduces this plan. I tried to write tests where I could. More could still be tested but its late and I'm running out of time.

It also cleans up some of the file fetching a little bit.

Example command to test:
```bash
cargo run -p orb-bidiff-squashfs-cli -- ota \
  --base ota://0.0.0-thebutlah.7176dea6+7176dea.2503182117.dev \
  --top ota://0.0.0-thebutlah.48fff015+48fff01.2503181807.dev \
  --out ~/P/patch \
  --download-dir ~/ota
```


TheButlah:
Tested manually, there was a small bug. I've pushed the changes to correct it and have now marked the PR as manually tested.

bidiff-squashfs/cli/src/lib.rs
```diff
@@ -11,6 +12,34 @@ pub async fn is_empty_dir(d: &Path) -> Result<bool> {
     Ok(tokio::fs::read_dir(d).await?.next_entry().await?.is_none())
 }
 
+trait PathExt {
```
alekseifedotov, line 15: Why is this needed? Why not call `tokio::fs::metadata` and then figure out that the objest is a directory and exists? 

Above there is code:
```
        assert!(d.is_dir_async().await, "{d:?} was not a directory");
        assert!(d.exists_async().await, "{d:?} does not exist");
```

Which will call `tokio::fs::metadata` twice, for no reason.

alekseifedotov (dismissed):
What is a use-case for diffing two OTAs with a CLI tool? I thought that the plan was to diff individual files and do that lazily on the server side.

bidiff-squashfs/cli/src/lib.rs
TheButlah: That's exactly what PathExt does, it's just convenience to be able to call that functionality using `var.func()` instead of `func(var)`

TheButlah: The asserts are redundant but that is intentional, I wanted to asset any preconditions.

TheButlah:
We need a tool we can run locally without CI, so that we can develop for the update agent. Not only that, getting the backend up would also take time.

My approach here is to optimize for local development, so that I can continue to test the update agent and finish it's implementation, and later incorporate this code into the backend.

bidiff-squashfs/cli/src/lib.rs
valff: Agree with @alekseifedotov, those methods are way too simple to justify introduction of an extension trait. For example `is_dir_async` can be written as:

```
fs::metadata(self).await.is_ok_and(|x| x.is_dir())
```

Which is IMO clearer than a new method.

And additionally I agree that this way we can reuse the Metadata object.

valff:
Is this an unfinished PR? Because `execute_plan` still has a `todo!()` call?

TheButlah:
These PRs are incremental, yes. I'm merging as I develop.

bidiff-squashfs/cli/src/lib.rs
TheButlah: That is significantly more boilerplate for no practical win. All the tests would need that boilerplate. This isn't high performance code.

The same mentality is used other places, for example I copy and allocate in many places to be more ergonomic and sacrifice performance. See the hashset and hashmap stuff for example.

I'll make the changes here, but overall it's important to contextualize the actual performance needs of the software.

TheButlah: fixed, I've deleted PathExt

valff (approved):

PR #445, opened 2025-03-26
feat(mcu-util): better ack error handling

fouge:
show when ack returns an error


TheButlah:
Don't forget to change the commit to something on main

PR #448, opened 2025-03-26
feat(location): FM-405 orbsoft location dbus status

paulquinn00:
Orb Location daemon
- use BackendStatus1 dbus to send location status


location/src/dbus.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use crate::data::NetworkInfo;
+
+pub struct BackendStatus {
+    backend_status_proxy: BackendStatusProxy<'static>,
+}
+
+impl BackendStatus {
+    pub async fn new() -> Result<Self> {
+        let connection = Connection::session().await?;
```
TheButlah, line 13: very important that the session can be passed in via dependency injection (i.e. provided as an argument, instead of constructed in the function). This allows the code to be more testable, and lets us set up a temporary dbus in tests, constructing the connection from it, and then passing that connection to the code. Without that, there is no way to independently test code, since it will all be sharing the same bus.

For an example of how this works, see [here](https://github.com/worldcoin/orb-software/blob/main/supervisor/tests/it/main.rs) and [here](https://github.com/worldcoin/orb-software/blob/main/supervisor/tests/it/helpers.rs)

Michael-Preibisch (approved):

PR #450, opened 2025-03-27
feat(bidiff-squashfs): generate new claim for patch

TheButlah:


TheButlah:
I've decided to create https://linear.app/worldcoin/issue/ORBS-415/patch-hash-and-size-info-in-claim as a separate issue/PR, so that we can merge this code and keep reviews small. Basically, the consequence is that the generated claim does not have the correct hash or size right now, but given that this PR still introduces a lot of important foundational work, I think we can mark as ready to review.

paulquinn00 (approved):

PR #452, opened 2025-03-27
feat(update-agent): remove deserialize, add serialize for Unchecked*

TheButlah:
I mistakenly added a Deserialize impl on the update agent Claim. This is a bad idea, because the update agent performs verification on its claim in order to check the manifest signature.

I've removed my impl and instead I will use the `UncheckedClaim` in bidiff-squashfs.

I've also added the ability to serialize an UncheckedClaim.

### DIFF
--- a/update-agent/core/src/claim.rs
+++ b/update-agent/core/src/claim.rs
@@ -207,7 +207,7 @@ impl<'a> Iterator for ComponentIter<'a> {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Debug)]
 pub struct Claim {
     version: String,
     manifest: crate::Manifest,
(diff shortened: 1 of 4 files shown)


update-agent/core/src/claim.rs
```diff
@@ -207,7 +207,7 @@ impl<'a> Iterator for ComponentIter<'a> {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug)]
```
alekseifedotov, line 210: Why does it need to serialize the claim? Does it save the claim somewhere?

update-agent/core/src/claim.rs
```diff
@@ -333,14 +348,14 @@ mod serde_imp {
...
-    pub(super) struct UncheckedClaim {
+    #[derive(Debug, Deserialize, Serialize)]
+    pub struct UncheckedClaim {
         version: String,
         manifest: UncheckedManifest,
         /// Signed sha256 hash of the claim
         #[serde(rename = "manifest-sig")]
         signature: Option<String>,
-        sources: HashMap<String, Source>,
+        pub sources: HashMap<String, Source>,
```
alekseifedotov, line 358: why this change? There is an access method for `sources`.

alekseifedotov (approved):

PR #459, opened 2025-04-02
feat(io-utils): implement ClampedSeek

TheButlah:
Needed for binary diffing. We need to clamp a Read + Seek within a particular range.

This will get used to replace the current `gpt::TakePartition` code in the update agent which is inherently bugged when doing SeekFrom::End

tbh writing the test cases was hell :dizzy_face:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5726,6 +5726,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "orb-io-utils"
+version = "0.0.0"
+
+[[package]]
 name = "orb-jwk-util"
 version = "0.0.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ members = [
   "gst-jpeg",
   "header-parsing",
   "hil",
+  "io-utils",
   "jwk-util",
   "location",
   "mcu-interface",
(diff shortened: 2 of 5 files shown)


dangirsh:
What Claude Code thinks:

1. **Interrupted Updates with Inconsistent State**: If an OTA update is interrupted (power loss, system crash), the `ClampedSeek` doesn't provide transaction-like guarantees. The reader position could be left in an inconsistent state between calls, leading to partial reads/writes that corrupt the rootfs. Since the implementation maintains separate cursor state, an interruption could leave the underlying stream and wrapper out of sync.

2. **SD Card Sector Boundary Misalignment**: SD cards operate on sector boundaries (typically 512 bytes). The implementation doesn't align reads to sector boundaries, which could cause issues if the OTA process assumes sector alignment. This might lead to partial sector writes that damage filesystem structures, particularly at partition boundaries where alignment is critical.

3. **Resource Exhaustion During Large Transfers**: During large file transfers common in OTA updates, the implementation doesn't provide buffering optimization or progress reporting. For a 64GB card, the system might exhaust memory or file handles while processing large rootfs components, especially if many small reads are performed instead of optimized block reads aligned with SD card hardware characteristics.

sri9311 (approved):

AlexKaravaev (approved):

PR #461, opened 2025-04-02
refactor(update-agent): get rid of gpt fork and TakePartition

TheButlah:
rebased on #459

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -744,7 +744,7 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "tracing",
- "uuid 1.11.0",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


io-utils/src/clamped.rs
```diff
@@ -0,0 +1,557 @@
...
+        );
+        self.cursor = from_self_start;
+
+        Ok(self.cursor)
+    }
+}
+
+impl<RS: Read + Seek> Read for ClampedSeek<RS> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        if self.cursor >= self.clamp.end {
```
alekseifedotov, line 109: I don't think this short-circuiting is needed, as it will be handled by a regular code path.

alekseifedotov (approved):
I'm glad that you removed the gpt for.
I am astounded by the amount of boilerplate to make ClampedSeek.

TheButlah:
marking as draft while I rework a bit more and test

TheButlah:
> I am astounded by the amount of boilerplate to make ClampedSeek.

Yeah, its about 128 lines for the actual struct implementation. However it could be half if I didn't check for overflows etc.

Most of the boilerplate actually is in the tests, not sure how to improve that. I'm also not happy with how verbose the tests are. I think it may be a skill issue on my end, or maybe seeking and reading is just super edge case prone.

io-utils/src/clamped.rs
TheButlah: it actually is necesssary otherwise we do an extraneous self.inner.read and/or panic on checked_sub

TheButlah:
I've run two OTAs now on my pearl orb via --update-location and everything seems to have functioned correctly.

PR #463, opened 2025-04-03
feat(ui): Pearl biometric-flow UX

karelispanagiotis:
1. Adds failure animation in case the processing phase returns an error. (red center LED & error sound).
2. Halting & Resuming the progress bar.
      * Halts the progress bar when the user is not in range (also the music stops playing).
      * Resumes the progress bar when the user enters the range, and the progress bar moves faster to catch-up with the "missed" progress. This is encouraging to the the user, as they did something positive (entered the correct range). (also the music playing resumes).

codebase changes: Introduces a new "composite" animation, which consists of the all the ux events that need to happen in a sequence during the signup (capturing and processing).


TheButlah:
label with tested:yes if you have manually tested this on your orb. If so I'll approve.

karelispanagiotis:
resolves O-3021

PR #464, opened 2025-04-04
fix(ci): update openssl

karelispanagiotis:
Fixes `cargo-deny` by updating the `openssl` dependency.


TheButlah:
we shouldn't even have an openssl dependency...

TheButlah (approved):

PR #466, opened 2025-04-04
WIP: fixes for orin slot switching

alekseifedotov:
Orin has somewhat differen way to confirm that the boot was successful after the slot switch. Make the changes to fix it for orin.

--
For reviewers: it is better to review each individual commit


TheButlah:
tbh I have no idea. @vmenge can you review? Otherwise ill approve to unblock.

TheButlah (approved):

PR #488, opened 2025-04-16
CriticalState UI

vmenge:
Shines the front ring purple in case of a critical Wifi issue that requires a power cycle.
This will be removed before the next event, just makes it easier to troubleshoot problems in EOL and field testing right now.

### DIFF
--- a/ui/src/engine/diamond.rs
+++ b/ui/src/engine/diamond.rs
@@ -1071,7 +1071,7 @@ impl EventHandler for Runner<DIAMOND_RING_LED_COUNT, DIAMOND_CENTER_LED_COUNT> {
                 self.set_ring(
                     LEVEL_FOREGROUND,
                     animations::Static::<DIAMOND_RING_LED_COUNT>::new(
-                        Argb::FULL_PURPLE,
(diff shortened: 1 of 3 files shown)


karelispanagiotis (approved):
LGTM, let's remove it before the event, or make the operator LED on the back of the orb convey this info.

Resolve conflicts and merge!

ui/src/engine/diamond.rs
```diff
@@ -1063,6 +1064,19 @@ impl EventHandler for Runner<DIAMOND_RING_LED_COUNT, DIAMOND_CENTER_LED_COUNT> {
...
             Event::VoiceOpenEyes => {}
+
+            Event::CriticalState {
+                state: CriticalState::WifiModuleNotInitialized,
+            } => {
+                self.set_ring(
+                    LEVEL_FOREGROUND,
+                    animations::Static::<DIAMOND_RING_LED_COUNT>::new(
+                        Argb::DIAMOND_FULL_PURPLE,
+                        Some(5.0),
```
fouge, line 1075: should the color lasts only 5 seconds? 🤔 what if it's missed by the operator?
you can pass None for infinite time

PR #490, opened 2025-04-17
Update agent loader

alekseifedotov:
This is a tool which downloads an ELF file from an URL, checks its signature and runs that ELF file. That is basically `wget && openssl dgst && exec` in a form of one rust package. 

The advantage of this rust binary over a script is in tighter security, this implementation does not store the file on the disk and downloads & checks the signature & runs the file **in memory**. Which mitigates the risk of the file being tampered with while on the disk.

The use for the tool is the follows: we release a base orb-platform version, where instead of `update-agent` we run this  loader  which tries to download the latest version of `update-agent` from github. That will allow us to change the `update-agent` version without doing a full OTA update. and maybe even hot-fix some bugs which otherwise break the OTA.

A short demo of the tool running on an orb is here: 

update-agent-loader

https://asciinema.org/a/aLkV2hKNPcX9Dd0eUHbBSWuK4

--

That same tool will work for `orb-core` too. In case if we need to release it often, we could, instead update the orb-core version on github releases and let orbs to download it.

--
P.S. the whole thing except for the test is vibe-coded while waiting for RTS & OTA on pearl.


update-agent-loader/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+rustc_private = true
+
+[features]
+default = []
+# The allow_http feature enables HTTP URLs for testing purposes
+# This feature cannot be enabled in release mode due to a compile-time check
+allow_http = []
+
+[dependencies]
+reqwest = { version = "0.11.4", features = ["blocking", "rustls-tls-native-roots"], default-features = false }
```
TheButlah, line 23: Use request from the security utils crate to ensure we are pinning certs properly

update-agent-loader/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+eyre = "0.6.9"
+tracing = "0.1.40"
+tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
+env_logger = "0.10.1"
+thiserror = "1.0.57"
+url = "2.2.2"
+nix = { version = "0.26", features = ["mman", "process"] }
+clap = { version = "4.4", features = ["derive"] }
+ed25519-dalek = "2.1.1"
+base64 = "0.22.1"
```
TheButlah, line 33: A lot of the above dependencies should be inherited

update-agent-loader/tests/functional_tests.rs
```diff
@@ -0,0 +1,124 @@
...
+    let secret_key_bytes = std::fs::read(secret_key_path)
+        .expect("Failed to read test secret key. Make sure to run in debug mode.");
+
+    SigningKey::from_bytes(secret_key_bytes.as_slice().try_into().unwrap())
+}
+
+/// Sign a binary with the specified keypair
+/// Returns a new Vec<u8> containing the original binary plus signature data
+fn sign_binary(data: &[u8], signing_key: &SigningKey) -> Vec<u8> {
+    // Create the signature
```
TheButlah, line 27: Should we just use ed25519ph with the signing context instead of concatenating a signing context?

And the signing context should be a const

update-agent-loader/tests/functional_tests.rs
```diff
@@ -0,0 +1,124 @@
...
+    signed_data.extend_from_slice(data);
+
+    // Add the signature
+    signed_data.extend_from_slice(&sig_bytes);
+
+    // Add the signature size (little-endian u32)
+    signed_data.extend_from_slice(&(sig_bytes.len() as u32).to_le_bytes());
+
+    // Add the magic bytes
+    signed_data.extend_from_slice(b"$WLD TO THE MOON");
```
TheButlah, line 46: Don't add Easter eggs in security critical code.

If you instead intended this to be a signing context like that which is used in ed25519ph, that is a good idea. But this code doesn't actually do that.

update-agent-loader/tests/functional_tests.rs
```diff
@@ -0,0 +1,124 @@
...
+    // Read the key from disk
+    let secret_key_bytes = std::fs::read(secret_key_path)
+        .expect("Failed to read test secret key. Make sure to run in debug mode.");
+
+    SigningKey::from_bytes(secret_key_bytes.as_slice().try_into().unwrap())
+}
+
+/// Sign a binary with the specified keypair
+/// Returns a new Vec<u8> containing the original binary plus signature data
+fn sign_binary(data: &[u8], signing_key: &SigningKey) -> Vec<u8> {
```
TheButlah, line 26: Newtype your sig for type safety. Ideally you would just be using the signature type from dalek.

update-agent-loader/src/main.rs
```diff
@@ -0,0 +1,43 @@
+use clap::Parser;
+use eyre::{eyre, Result};
+use tracing_subscriber::EnvFilter;
+use url::Url;
+
+/// Update agent loader that downloads and executes a binary from a URL
+#[derive(Parser, Debug)]
+#[clap(author, version, about, trailing_var_arg = true)]
```
TheButlah, line 8: Use same convention as rest of repo, you are missing build info, etc

update-agent-loader/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// Arguments to pass to the downloaded executable
+    /// All arguments after -- will be passed to the executable
+    #[clap(last = true)]
+    exec_args: Vec<String>,
+}
+
+fn main() -> Result<()> {
+    // Initialize tracing with env_logger compatibility
+    tracing_subscriber::fmt()
```
TheButlah, line 22: Use orb-telementary

update-agent-loader/tests/functional_tests.rs
```diff
@@ -0,0 +1,124 @@
...
+fn load_test_signing_key() -> SigningKey {
+    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+    let secret_key_path = std::path::Path::new(&manifest_dir)
+        .join("target")
+        .join("debug")
+        .join("keys")
+        .join("secret_key.bin");
+
+    // Read the key from disk
+    let secret_key_bytes = std::fs::read(secret_key_path)
```
TheButlah, line 18: It's a test, so you can generate the signing key instead of needing a secret the environment has to provide

update-agent-loader/tests/functional_tests.rs
```diff
@@ -0,0 +1,124 @@
...
+    let echo_binary = std::fs::read("/bin/echo").expect("Failed to read /bin/echo");
+    let signed_binary = sign_binary(&echo_binary, &signing_key);
+
+    // Create the mock endpoint serving the signed binary
+    rt.block_on(
+        Mock::given(method("GET"))
+            .and(path("/binary"))
+            .respond_with(
+                ResponseTemplate::new(200)
+                    .set_body_raw(signed_binary, "application/octet-stream"),
```
TheButlah, line 87: Just use set_body_bytes

update-agent-loader/sign.py
```diff
@@ -0,0 +1,70 @@
...
+    # Write output
+    with open(args.output_file, "wb") as f:
+        f.write(original_data)
+        f.write(signature)
+        f.write(sig_len.to_bytes(4, "little"))
+        f.write(WLD_TAG)
+
+    print(f"Signed {'(replaced existing signature)' if already_signed else ''}and saved to: {args.output_file}")
+
+if __name__ == "__main__":
```
TheButlah, line 69: You already have rust code to verify. Why not use use dalek to sign? Having two different languages for the two sides of a cryptographic operation seems to be asking for trouble.

update-agent-loader/sign.py
```diff
@@ -0,0 +1,70 @@
...
+import argparse
+from cryptography.hazmat.primitives.asymmetric import ed25519
+
+WLD_TAG = b"$WLD TO THE MOON"
+TAG_LEN = len(WLD_TAG)
+SIG_LEN_SIZE = 4  # 4 bytes for signature length
+
+def load_raw_ed25519_private_key(path):
+    with open(path, "rb") as f:
+        raw_key = f.read()
```
TheButlah, line 13: Why not store it how we stored other signing keys? For example we use jwk to load keys in the update agent for the manifest sig.

See orb security utils.

Also key import stuff should go in orb security utils so that we can ensure we have one spot to pin the pubkey

update-agent-loader/sign.py
```diff
@@ -0,0 +1,70 @@
...
+    if len(raw_key) != 32:
+        raise ValueError("Expected 32-byte Ed25519 private key")
+    return ed25519.Ed25519PrivateKey.from_private_bytes(raw_key)
+
+def read_and_strip_existing_signature(data):
+    """If data ends with WLD_TAG and signature length, strip them."""
+    if len(data) < TAG_LEN + SIG_LEN_SIZE:
+        return data, False
+
+    if data[-TAG_LEN:] != WLD_TAG:
```
TheButlah, line 23: This sort of manual binary parsing is the type of spot that bugs get introduced in security critical code.

Id eliminate the magic string, stick to known key formats, and use rust not python.

TheButlah (changes requested):

TheButlah (commented):
Haven't reviewed lib.rs yet because there are so many issues already in the rest of the code.

At some point, when the review takes longer than just writing the code yourself, I have to question if the vibe coding is worth it.

update-agent-loader/src/lib.rs
```diff
@@ -0,0 +1,532 @@
...
+impl MemFileState for Verified {}
+
+/// A memory-mapped view of a MemFile that implements Deref to [u8]
+pub struct MemFileMMap<'a> {
+    mapped_ptr: *mut c_void,
+    size: usize,
+    _file: &'a MemFile<Unverified>, // Keep a reference to the file to ensure it lives as long as the mapping
+}
+
+impl<'a> MemFileMMap<'a> {
```
TheButlah, line 123: for the code that creates a secure, in memory temp file. I would prefer if we break that into its own crate. It is both generally useful (I think we tried to do this with the key retrieval crate?) and also hard to test and review when its bundled up with the rest of this code.

Given that its a fairly low level and also unsafe primitive, its well suited to be its own crate. what you name it, idc, name it orb-foobar for all I care XD

TheButlah (changes requested):

update-agent-loader/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+
+# Enable unstable features needed for compilation
+[package.metadata.rust-analyzer]
+rustc_private = true
+
+[features]
+default = []
+# The allow_http feature enables HTTP URLs for testing purposes
+# This feature cannot be enabled in release mode due to a compile-time check
+allow_http = []
```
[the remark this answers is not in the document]
alekseifedotov: How would you suggest to test it with wiremock?

update-agent-loader/tests/functional_tests.rs
alekseifedotov: But the pubkey is embedded in the binary, how would I inject the generated key in the binary?

update-agent-loader/sign.py
alekseifedotov: >Having two different languages for the two sides of a cryptographic operation seems to be asking for trouble.

That is done exactly to avoid troubles. The idea is to make signing and verification more dissimilar so it will be harder to evolve the codebase in such way that it weakens security.

I choose python as second language because it does not require installing crazy dependencies or building anything to sign the binary.

update-agent-loader/sign.py
alekseifedotov: > Why not store it how we stored other signing keys? For example we use jwk to load keys in the update agent for the manifest sig.

That is an option. I could store it in a PEM file.

update-agent-loader/src/lib.rs
```diff
@@ -0,0 +1,532 @@
...
+    env,
+    ffi::{c_void, CString},
+    io::{self, Write},
+    num::NonZeroUsize,
+    ops::Deref,
+    os::fd::{AsRawFd, FromRawFd, OwnedFd, RawFd},
+};
+
+use ed25519_dalek::{Signature, Verifier, VerifyingKey};
+use nix::{
```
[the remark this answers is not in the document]
alekseifedotov: 2 questions:
 - why rustix is better than nix?
 - why rust standart library does not provide such basic functionality?

update-agent-loader/src/lib.rs
```diff
@@ -0,0 +1,532 @@
...
+}
+
+#[derive(Debug)]
+pub enum MemFileError {
+    MemfdCreate(Errno),
+    Io(io::Error),
+    Mmap(nix::Error),
+    Fstat(nix::Error),
+    /// Failed to parse or verify signature
+    SignatureError(String),
```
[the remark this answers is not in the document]
alekseifedotov: what is error modality? Maybe it is duplicated, the whole error handling is very repetitive.

update-agent-loader/src/lib.rs
```diff
@@ -0,0 +1,532 @@
...
+        let signature = Signature::from_bytes(&sig_bytes);
+
+        // Parse the public key (using the hardcoded key for now)
+        // Load the hardcoded public key
+        let public_key = VerifyingKey::from_bytes(PUBLIC_KEY_BYTES)
+            .map_err(|e| MemFileError::SignatureError(format!("Failed to parse public key: {}", e)))?;
+
+        // The data to verify is everything except the signature, its size and magic bytes
+        let data = &mmap[..mmap.len() - sig_size - FOOTER_SIZE];
+
```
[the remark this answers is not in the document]
alekseifedotov: It is already separated. there is some unsafe in MemFileMMap and MemFile and they don't touch cryptography.

update-agent-loader/src/lib.rs
alekseifedotov: How are you even planning to re-use it? It is super-specific to this use-case.

If someone wants to create a memfd file, they could just call one function `memfd_creat()` or `memfd_secret()`

update-agent-loader/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "update-agent-loader"
+version = "0.1.0"
+edition = "2021"
+authors = ["Worldcoin <engineering@worldcoin.org>"]
+description = "A loader for the update-agent"
+license = "MIT OR Apache-2.0"
```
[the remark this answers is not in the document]
alekseifedotov: done

update-agent-loader/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "update-agent-loader"
+version = "0.1.0"
```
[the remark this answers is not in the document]
alekseifedotov: done

update-agent-loader/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+edition = "2021"
+authors = ["Worldcoin <engineering@worldcoin.org>"]
```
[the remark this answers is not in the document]
alekseifedotov: done

update-agent-loader/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+        .with_env_filter(EnvFilter::from_default_env())
+        .init();
+
+    // Parse command line arguments
+    let args = Args::parse();
+
+    // Use provided URL or fallback to default
+    let url_str = args
+        .url
+        .unwrap_or_else(|| "https://example.com/path/to/executable".to_string());
```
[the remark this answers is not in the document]
alekseifedotov: removed

update-agent-loader/src/lib.rs
```diff
@@ -0,0 +1,532 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum DownloadError {
+    #[error("HTTP client error: {0}")]
+    ClientError(#[from] reqwest::Error),
+
+    #[error("I/O error: {0}")]
+    Io(#[from] io::Error),
+
+    #[error("Memory file error: {0}")]
```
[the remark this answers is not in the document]
alekseifedotov: changed it to `MemFD`

update-agent-loader/sign.py
TheButlah: can you elaborate on why two entirely separate implementations helps avoid evolving the code to weaken security?

update-agent-loader/tests/functional_tests.rs
TheButlah: have a `run` helper, main calls run, main hard codes pubkey.

test `run` instead of main.

update-agent-loader/tests/functional_tests.rs
TheButlah: you can disregard this - a signing context still makes good sense to do, but this code is for magic strings.

update-agent-loader/tests/functional_tests.rs
TheButlah: after more discussion, it was raised that it really was intended as a magic string.

IMO i really think that a separate .sig file makes more sense. But if you have a strong opinion here, I guess it would be ok to concatenate it at the end of the file.

TheButlah (changes requested):

update-agent-loader/Cargo.toml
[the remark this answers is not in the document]
TheButlah: add a config struct or some other arguments to `run`. `main` calls run with hard coded args (use this pubkey, force https, call this closure for the exec). Move the bulk of your code out of `main` and into `run`.

Now your tests can control the pubkey, control if https is used, and avoid the actual fork+exec. it makes it a lot more testable.

update-agent-loader/sign.py
TheButlah: PEM or a `jwk` both work. although in the update agent we used `JWK` so probably a `JWK` is best.

TheButlah:
discussed on call with @alekseifedotov we will seek to merge ASAP as soon as CI passes

Dorex45 (approved):
ACK

TheButlah (approved):
Yolo

PR #492, opened 2025-04-17
feat: add boster-leds keep-on mode for mcu-util and set brightenss th…

AlexKaravaev:
…rough cli.

Also existing booster command in mcu-util was not working for me.


mcu-util/src/orb/main_board.rs
```diff
@@ -366,43 +333,93 @@ impl MainBoard {
                 .await
             {
                 Ok(CommonAckError::Success) => {
-                    info!("LEDs disabled");
+                    info!("White/booster LEDs turned off");
```
Qbicz, line 336: so booster & white are the same LEDs on Diamond?

AlexKaravaev: Yes, I think just the naming is a bit confusing. In control-api we call them "white leds", it's the same leds. But also at the same time rgbleds can be just white.

PR #493, opened 2025-04-17
feat(hil): allow user to pass --timeout to login command

Qbicz:
Configurable timeout is useful for many orb-hil use cases that require fast check of orb presence.


TheButlah:
fix ci plz :pray:

PR #494, opened 2025-04-21
feat(hil): implement readout protection disable

TheButlah:


TheButlah:
Also I *think* but have not tested, that this will work on any hardware probe that probe-rs supports.

alekseifedotov (approved):
That is a really nice feature, would be actually useful for developers to use this tool as a standalone tool.

PR #496, opened 2025-04-22
fix(backend-status): set it to actually restart

paulquinn00:


backend-status/debian/orb-backend-status.worldcoin-backend-status.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Worldcoin Backend Status
-Requires=worldcoin-backend-online.target
```
TheButlah, line 3: shouldnt it *also* need backend-online?

TheButlah (approved):

backend-status/debian/orb-backend-status.worldcoin-backend-status.service
paulquinn00: you're right

PR #497, opened 2025-04-23
fix: sound volume on diamond

AlexKaravaev:


karelispanagiotis:
I thought that the `SoundVolume` and other events where handled one function call above this (what I mention happens for Pearl).

I am wondering why rust compiler does not complain about not covering all of the enum variants.

karelispanagiotis (approved):
LGTM.
We probably need to refactor this whole binary.

PR #499, opened 2025-04-23
feat(bidiff-cli): implement upload to s3

TheButlah:
vibe coded with codex + o3

### DIFF
--- a/nix/machines/home-hil.nix
+++ b/nix/machines/home-hil.nix
@@ -55,8 +55,6 @@ in
   };
   programs.zellij = {
     enable = true;
-    enableBashIntegration = true;
-    enableZshIntegration = true;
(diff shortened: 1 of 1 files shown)


s3-helpers/src/upload_dir.rs
```diff
@@ -0,0 +1,108 @@
...
+                files.push(path);
+            }
+        }
+    }
+
+    Ok(files)
+}
+
+// Integration tests for this module are located in `s3-helpers/tests/upload_dir.rs` so that they
+// can share the LocalStack infrastructure with the other S3 helper tests.
```
TheButlah, line 108:
```suggestion
// can share the LocalStack infrastructure with the other S3 helper tests.
```

PR #516, opened 2025-04-30
feat(backend-status): Collect orb name and jabil id in status

paulquinn00:
- Collect orb name and jabil id in status to send to orb backend

So Fleet Commander can search by these

### DIFF
--- a/backend-status/src/backend/status.rs
+++ b/backend-status/src/backend/status.rs
@@ -1,7 +1,7 @@
 use chrono::Utc;
 use eyre::Result;
 use orb_endpoints::{v2::Endpoints as EndpointsV2, Backend};
-use orb_info::OrbId;
+use orb_info::{OrbId, OrbJabilId, OrbName};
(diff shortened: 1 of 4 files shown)


andronat (approved):

paulquinn00:
> In general ok for me, but I think backend should be our single source of truth - /usr/persistence on orb can be changed/messed up manually(and was frequently messed in the past), so I think it would be better if we map orb-id to jabil-id somehow on our backend instead of always trusting the orb. What do you think? Also tagging @sfikastheo because we had a lot of discussions about nuking jabil-id/orb-name completely from an orb and rather having a service that pulls this info from backend dynamically.
> 
> Understand though that ^ will require more work and having this pr as a stopgap is fine.

It's a good point, but the orb should be the SoT, not any backend. It has the final say in what's burned into it.
As for jabil id being changeable, what are we stamping on the case for Diamond?

PR #519, opened 2025-05-09
refactor(orb-slot-ctrl): Move efivar to it's own crate

pophilpo:
1. Preparing for larger refactor, that will allow `orb-slot-ctrl` to work between pearl and diamond.

2. Moved `efivar` to it's own crate because `efivar` is being used outside the `orb-slot-ctrl ` (e.g. capsule-updates) and that usage does not pertain to `orb-slot-ctrl`.

( Co-authored by @vmenge )


efivar/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        Ok(Self { path })
+    }
+
+    pub fn get_var(
+        &self,
+        relative_path: impl AsRef<Path>,
+    ) -> Result<EfiVar, EfiVarDbErr> {
+        let relative_path = relative_path.as_ref();
+        if relative_path.is_absolute() {
```
alekseifedotov, line 50: If that is an attempt to protect agains an attacker using absolute path to write an aribitrary file, then you should also check that the path does not have `../` in it. 

Ideally, you would want to check that the path has no `/` in it.

efivar/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn read(&self) -> Result<Vec<u8>> {
+        fs::read(&self.path).wrap_err_with(|| format!("Failed to read {:?}", self.path))
+    }
+
+    /// This function will create a efi file if it does not exist,
+    /// Will entirely replace its contents if it does.
+    pub fn write(&self, buffer: &[u8]) -> Result<()> {
+        let err = |msg| move || format!("Failed to {msg} {:?}", self.path);
+
+        let file_exists = fs::exists(&self.path)?;
```
alekseifedotov, line 82: That is TOCTOU bug: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

efivar/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+            let new_attributes = original_attributes & !ioctl::IMMUTABLE_MASK;
+            ioctl::write_file_attributes(&file_read, new_attributes)
+                .wrap_err_with(err("make file mutalbe"))?;
+
+            fs::write(&self.path, buffer).wrap_err_with(err("write to file"))?;
+            // Make file immutable again.
+            ioctl::write_file_attributes(&file_read, original_attributes)
+                .wrap_err_with(err("make file immutable"))
+        } else {
+            fs::write(&self.path, buffer).wrap_err_with(err("write to file"))
```
alekseifedotov, line 100: If the file does not exist, it will be created, but won't be made immutable, that is different behavior compared to when the file does exist and modified.

alekseifedotov (changes requested):

efivar/src/lib.rs
vmenge: This is the same behaviour as before with the `create_and_write` function though (we essentially merged `create_and_write` and `write` to have an API similar to `std::fs` and reduce API surface). We thought about the same thing but I asked @pophilpo to keep it like this to preserve functionality.

Do you know if there was a reason `create_and_write` wasn't marking it as immutable before? Otherwise we can add the change to make it immutable :+1:

efivar/src/lib.rs
vmenge: Removed the `fs::exists` to reduce the extra `open` and not increase chance of TOCTOU :+1:

efivar/src/lib.rs
vmenge: Ah no no, this is just here to avoid people from trying to get efivars through absolute paths (would make it harder to test and go against the `EfiVarDb` abstraction)

PR #525, opened 2025-05-26
refactor(backend): added build type enum to better distinguish between build configuration

bsayed:
Introduces a `BuildType` enum to better distinguish between build configurations in backend selection logic.


AlexKaravaev:
could we reuse https://github.com/worldcoin/orb-software/blob/main/orb-info/src/orb_os_release.rs here?

bsayed:
> could we reuse https://github.com/worldcoin/orb-software/blob/main/orb-info/src/orb_os_release.rs here?

In theory, we can if we modify the `OrbRelease` enum in this [file](https://github.com/worldcoin/orb-software/blob/main/orb-info/src/orb_os_release.rs) to add a new release type, possibily `Anaysis`, but we still have to modify the `Backend::from_env_or_build_type()` function and add a dependency on the `orb_os_release` feature in the `orb-endpoint`  project. Let me know if you prefer this solution.

AlexKaravaev:
I think its a bit cleaner, but I dont have strong feelings - as you wish

AlexKaravaev (approved):

PR #529, opened 2025-05-28
fix(update-agent): working binary for both Pearl and Diamond platforms

vmenge:
Tested for:
Diamond + Pearl (with capsule)


update-agent/debian/worldcoin-update-agent.service
```diff
@@ -1,19 +1,16 @@
...
+Wants=worldcoin-supervisor.service mnt-scratch.mount mnt-updates.mount
+Requires=usr-persistent.mount worldcoin-backend-online.target
 
 [Service]
 # As update-agent has no dependencies, it should be fine to make this Type=simple (rather than
 # Type=oneshot. This way, RestartForceExitStatus= works.
 Type=simple
 SyslogIdentifier=worldcoin-update-agent
 Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/worldcoin_bus_socket
-ExecStartPre=/usr/local/bin/orb-ensure-mnt-updates-writable
```
AlexKaravaev, line 16: Not needed anymore for Pearl? or updates are very rarely unwritable?

update-agent/src/component.rs
```diff
@@ -273,6 +274,38 @@ impl Component {
...
+                .on_disk
+                .to_str()
+                .wrap_err("failed to get sec_mcu bin path")?;
+
+            Command::new("orb-mcu-util")
+                .args(["--can-fd", "image", "update", "--path", file, "security"])
+                .stdout(Stdio::inherit())
+                .stderr(Stdio::inherit())
+                .status()
+                .wrap_err("Failed to update main mcu using orb-mcu-util")?;
```
AlexKaravaev, line 304: not main mcu
```suggestion
                .wrap_err("Failed to update sec mcu using orb-mcu-util")?;
```

update-agent/src/main.rs
```diff
@@ -663,47 +668,35 @@ fn finalize_normal_update(
...
-                }
-            }
-            Err(_) => {
-                warn!("Capsule update was not detected");
-            }
+    if let Ok(data) = EfiVarDb::from_rootfs("/")
+        .and_then(|db| db.get_var(EFI_OS_INDICATIONS))
+        .and_then(|var| var.read())
+    {
+        if data.value() == &EFI_OS_REQUEST_CAPSULE_UPDATE[4..] {
```
AlexKaravaev, line 684: I needed to look up why we need to do [4..], maybe we need to leave a comment?

update-agent/src/main.rs
```diff
@@ -663,47 +668,35 @@ fn finalize_normal_update(
...
+            slot_ctrl
+                .mark_slot_ok(target_slot.into())
```
AlexKaravaev, line 686: why?

update-agent/t/orb_update_agent.conf
```diff
@@ -3,8 +3,8 @@ components = "/usr/persistent/components.json"
 cacert = "/etc/ssl/worldcoin-staging-ota.pem"
 clientkey = "/etc/ssl/private/worldcoin-staging-ota-identity.key"
 update_location = "/mnt/claim.json"
-workspace = "/mnt/updates"
-downloads = "/mnt/updates/downloads"
+workspace = "/mnt/scratch"
```
AlexKaravaev, line 6: will this be ok for pearl too?

AlexKaravaev: ah wait, what is 't' directory? is it tests? Should it be just named tests? why it is named like this

AlexKaravaev (changes requested):

update-agent/debian/worldcoin-update-agent.service
sfikastheo: Never worked reliably in the first place IIRC

update-agent/src/main.rs
sfikastheo: You are returning early, UEFI will handle the slot-switching. This is unchanged logic

AlexKaravaev: but we were not setting the slot as ok previously when we were doing capsule update - we were setting counter to max, but not in this specific code path

AlexKaravaev: I think it's something new, I don't see anything bad with it though

PR #538, opened 2025-06-10
fix(mcu-util): recognize dvt

fouge:
fix main board detection


AlexKaravaev (approved):

AlexKaravaev:
we should release and bump orb-os after

PR #548, opened 2025-06-24
feat(agent-iroh): initial commit

TheButlah:
Implements:

* The first ever orb-core agent that doesn't live in the orb-core repo (dogfooding future potential for open sourcing agents)
* Agent manages iroh endpoint (this is essentially just QUIC, which uses a UDP socket + multiplexed protocols via [ALPN](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation))
* messages to agent allow for connecting to peers, accepting incoming connections, and disabling connections
* Builder structs to fully configure most important iroh settings
* Demo application in the form of the examples directory - `broker.rs` is a minimal orb-core broker, and `phone.rs` is pretending to be the phone.
* When the connection is migrated to a direct one, messages are logged to the console (see [ConnectionType](https://docs.rs/iroh/0.35.0/iroh/endpoint/enum.ConnectionType.html)). 

Latency numbers look very promising, but I need to test on a wider variety of devices, especially on machines that are remote to me.

Next steps once this PR is merged:
* update to latest iroh crate version, the devs added a bunch of things I asked for to make my life easier. They are super responsive!!
* add `orb-agent-iroh` as a dependency of `orb-core` under a feature flag.
* Port the example plans and functions on the broker to orb-core's plan.
* *for now* hard code the "phone's" pubkey into the UserDataBackend struct. This is just to simulate getting the pubkey from the QR code data.
* Use this pubkey to connect to the "phone" (which is for now a CLI app, later we can actually expose this as an android/ios app).
* forward the preflight checks through the established connection.
* add datadog metrics for latency numbers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ result
 
 # git worktrees
 /w
+
+# Produced sometimes
+.intentionally-empty-file.o
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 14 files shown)


TheButlah:
Yes this workstream is a PoC. However if fits our needs, it can be shipped later down the line.

AlexKaravaev (approved):

PR #551, opened 2025-06-24
chore: upgrade to edition 2024

TheButlah:


location/src/config.rs
```diff
@@ -142,26 +139,26 @@ impl Config {
...
-        } else {
-            warn!("ORB_BACKEND not set. Backend operations may fail.");
-            return Err(LocationError::ConfigError(
-                "ORB_BACKEND environment variable not set. Use the backend option or set environment variable.".into()
-            ));
-        }
-
-        Ok(())
-    }
+    // /// Set the environment variables based on the configuration
```
AlexKaravaev: i am not sure it's used. i grepped for setup_environment. I think it's better to delete the code, instead of commenting it out, but also I think we should put package maintainers into cc

seek-camera/wrapper/src/manager/mod.rs
```diff
@@ -254,7 +257,10 @@ mod tests {
         #[test]
         fn test_double_set_callback() {
             let dir = tempfile::tempdir().unwrap();
+            // TODO: Fix this shit
+            unsafe {
             std::env::set_var("SEEKTHERMAL_ROOT", dir.path());
```
AlexKaravaev: nit identation

ui/src/engine/operator/signup_phase.rs
```diff
@@ -270,7 +270,7 @@ impl Animation for SignupPhase {
                     .iter()
                     .enumerate()
                     .rev()
-                    .filter(|(_, &c)| !c.is_off())
+                    .filter(|(_, c)| !c.is_off())
```
AlexKaravaev, line 273: why we need this change?

TheButlah: Either it had to do with pattern matching changing a bit in 2024 edition: https://doc.rust-lang.org/edition-guide/rust-2024/match-ergonomics.html
or it was just always redundant.

location/src/config.rs
TheButlah: Just FYI @paulquinn00 all the env::set_var stuff we did in orb-info and orb-location is unsafe as of Rust 2024 (it technically, always was), and we should find a new path forward that doesn't rely on process-level global state.

TheButlah: Re: 
> I think it's better to delete the code, instead of commenting it out

Sure, I can do that

AlexKaravaev (approved):

PR #553, opened 2025-06-26
feat(agentwire): use quote_spanned!() for agent attributes

TheButlah:


valff (approved):
This is definitely more ergonomic, thanks!

PR #565, opened 2025-07-11
feat(orb-mcu-util): Allow passing FPS value for camera trigger

Qbicz:
You can now call it as:
orb-mcu-util optics trigger-camera face 25


fouge (approved):
🎉

PR #566, opened 2025-07-11
fix(hil): added lz4c symlink to lz4 binary

TheButlah:


TheButlah:
marking as tested: `nix run github:worldcoin/orb-software#tegraBash works, and the lz4c symlink exists.

I did not yet test if this resolves the flashing in the jeston scripts though.

Qbicz (approved):
Looks good!

TheButlah:
Marked as draft, the change causes too much to rebuild from source

PR #570, opened 2025-07-20
feat(orb-registration): rewrite Orb Registration Scripts in Python with Unified Support for Pearl and Diamond

pophilpo:
# Rewrite Orb Registration Scripts in Python with Unified Support for Pearl and Diamond

## 📝 Summary

This PR introduces a complete Python rewrite of the legacy shell scripts `gen-orb-id.sh` and `register-mongo.sh` into a single, dependency-free script: `orb-registration.py`.

### Unified Workflow
- Combines `gen-orb-id.sh` and `register-mongo.sh` functionality into one script
- Supports **both Pearl** and **Diamond** orb platforms with platform-specific workflows

### Python Standard Library Only
- Zero python dependencies
- Compatible with Python 3.6+


scripts/orb-registration-script/README.md
```diff
@@ -0,0 +1,362 @@
+# Orb Registration Script
```
AlexKaravaev: Let's delete all other scripts which are not needed anymore now for same purpose

scripts/orb-registration-script/orb-registration.py
```diff
@@ -0,0 +1,754 @@
...
+        self.build_dir = self.script_dir / "build"
+        self.artifacts_dir = self.script_dir / "artifacts"
+        self.core_app_url = "https://api.operator.worldcoin.org/v1/graphql"
+        self.persistent_size = 1024 * 1024
+        self.persistent_journaled_size = 10 * 1024 * 1024
+
+        # Configure backend
+        if args.backend == "stage":
+            self.domain = "https://management.internal.stage.orb.worldcoin.dev"
+            self.channel = "internal-testing"  # Default for stage
```
AlexKaravaev: dev_diamond_channel for diamond and internal-testing for pearls

scripts/orb-registration-script/orb-registration.py
```diff
@@ -0,0 +1,754 @@
...
+
+        for orb_id in orb_ids:
+            self.logger.info(f"Processing Diamond Orb ID: {orb_id}")
+            orb_name = self.register_orb_mongo(orb_id, cf_token, platform)
+            self.register_orb_core_app(orb_id, orb_name)
+            self.logger.info(f"Successfully processed Diamond Orb: {orb_id}")
+
+    def process_diamond_orb_pairs(self, orb_pairs: List[Tuple[str, str]]):
+        """Process Diamond orb ID+name pairs (register directly in Core-App)."""
+        for orb_id, orb_name in orb_pairs:
```
AlexKaravaev: We discussed it already but i don't see it anywhere - orb id is max 8 chars and if provided one is less you should pad with zeros and print a warning. if more than 8 - raise error

scripts/orb-registration-script/orb-registration.py
```diff
@@ -0,0 +1,754 @@
...
+                )
+                subprocess.run(
+                    ["setfacl", "-d", "-m", "u::rwx,g::rwx,o::rx", str(mount_point)],
+                    check=True,
+                )
+                subprocess.run(["sync"], check=True)
+            finally:
+                subprocess.run(["umount", str(mount_point)], check=True)
+
+    def register_orb_mongo(self, orb_id: str, cf_token: str, platform: str) -> str:
```
AlexKaravaev: We also discussed it already, but if orb is already registered, you could get details of an orb and return the name. Now you just fail

scripts/orb-registration-script/orb-registration.py
```diff
@@ -0,0 +1,754 @@
...
+        """Register orb in MongoDB and return orb_name."""
+        self.logger.info(f"Creating Orb record in Management API for Orb ID: {orb_id}")
+
+        data = {
+            "BuildVersion": self.args.hardware_version,
+            "ManufacturerName": "TFH_Jabil",
```
AlexKaravaev: Make this one as an argument and TFH_Jabil as default one

scripts/orb-registration-script/orb-registration.py
```diff
@@ -0,0 +1,754 @@
...
+            self.domain = "https://management.internal.stage.orb.worldcoin.dev"
+            self.channel = "internal-testing"  # Default for stage
+        elif args.backend == "prod":
+            self.domain = "https://management.internal.orb.worldcoin.dev"
+            self.channel = args.channel  # Use user-provided channel for prod
+        else:
+            raise ValueError(f"Invalid backend: {args.backend}")
+
+    def get_cloudflared_token(self) -> str:
+        """Get Cloudflare access token for the domain."""
```
AlexKaravaev: this is a nit: i think it's fine not to write any docstrings, if they just duplicate method name. no need to change anything

scripts/orb-registration-script/orb-registration.py
```diff
@@ -0,0 +1,754 @@
...
+    )
+    parser.add_argument(
+        "--mongo-token", help="MongoDB bearer token (overrides environment)"
+    )
+    parser.add_argument(
+        "--core-token", help="Core-App bearer token (overrides environment)"
+    )
+    parser.add_argument(
+        "--channel",
+        default="general",
```
AlexKaravaev: for prod, let's not set a default channel and hence if channel not set don't set it when you process orb ids.

In theory we could set different channels based on platform, but I think we first need to clean them up

scripts/orb-registration-script/orb-registration.py
```diff
@@ -0,0 +1,754 @@
...
+        orb_name = self.register_orb_mongo(orb_id, cf_token, platform)
+        self.set_orb_channel(orb_id, cf_token)
+        token = self.get_orb_token(orb_id, cf_token)
+
+        self.save_orb_artifacts(orb_id, orb_name, token, mount_point)
+        self.register_orb_core_app(orb_id, orb_name)
+
+        return orb_id
+
+    def process_diamond_orb_ids(self, orb_ids: List[str], cf_token: str):
```
AlexKaravaev: this is not a good design - you duplicate code both for platforms and also both for slightly different input args formats(e.g pairs vs ids), probably something very simple like base class would reduce some code duplication. but i comment here only just fyi and learning, i don't think we should spend a lot of time on this util script

scripts/orb-registration-script/README.md
```diff
@@ -0,0 +1,362 @@
...
+
+A python script for generating and registering Orb devices across both Pearl and Diamond platforms with MongoDB and Core-App.
```
AlexKaravaev: In general I think documentation is great. However, I feel like most of this particular readme can be comments in code or code itself. Maybe that's a personal preference but I think code ideally (where it can) should document itself. we can have documentation on how to set up things, but most of the points in this README is just unnecessary.

I would try to make this README much smaller and remove obvious sections. I would also probably move it to [docs](https://github.com/worldcoin/orb-software/tree/main/docs)?

AlexKaravaev (changes requested):

scripts/orb-registration-script/README.md
pophilpo: I just think in general when using a script that you don't really touch (not working on developing it) you tend to just stick with reamde and run the script as  is, without diving into the code. I can improve on this as you mentioned though

scripts/orb-registration-script/README.md
AlexKaravaev: nit but i would change directory name: it's already under scripts, no need for double -script postfix here

scripts/orb-registration-script/README.md
AlexKaravaev: shoould it still be moved to docs?

PR #571, opened 2025-07-20
feat: extend update-agent reporting

AlexKaravaev:
Extend update agent reporting, modify backend status connect to not depend on if update agent is alive or not. Example of status sent with new update and new field "state"
```
Jul 20 14:06:21 localhost.localdomain worldcoin-backend-status[627981]: 
Sending status to backend: CurrentStatus 
....update_progress: Some(UpdateProgress { download_progress: 100, processed_progress: 100, install_progress: 100, total_progress: 100, error: None, state: NoNewVersion }), ...
```


backend-status/src/dbus/intf_impl.rs
```diff
@@ -148,6 +188,8 @@ impl BackendStatusImpl {
             "Updating backend-status"
         );
 
+        info!("Sending status to backend: {:?}", current_status);
```
Michael-Preibisch: Are you sure we want to log the entire status struct?

backend-status/src/dbus/intf_impl.rs
```diff
@@ -161,8 +203,23 @@ impl BackendStatusImpl {
     }
 
     fn get_available_status(&self) -> Option<CurrentStatus> {
+        let should_force_send =
+            if let Ok(mut force_flag) = self.force_immediate_send.lock() {
+                let force = *force_flag;
+                if force {
+                    *force_flag = false; // Reset the flag after checking
+                    info!(
```
Michael-Preibisch: Nit: Weird formatting

backend-status/src/update_progress.rs
```diff
@@ -18,132 +20,297 @@ pub enum UpdateProgressErr {
...
-        })
-        .map(|c| {
-            if c.state == ComponentState::Downloading {
-                c.progress as u64
+
+    fn extract_progress_data(
+        changed_properties: &std::collections::HashMap<&str, Value<'_>>,
+    ) -> Result<Option<(Option<u8>, UpdateAgentState)>, UpdateProgressErr> {
+        let overall_progress =
+            if let Some(progress_value) = changed_properties.get("OverallProgress") {
```
Michael-Preibisch: Here and in many other places, we use 'magic strings' in code. I think we should define them as constants.

update-agent/src/claim.rs
```diff
@@ -143,7 +145,12 @@ fn from_remote(
                 Cow::Borrowed("")
             }
         };
-        Err(Error::status_code(status, msg))
+
+        if status == StatusCode::NOT_FOUND && msg.contains("no new version") {
```
Michael-Preibisch: Depending on message returned from the backend ("no new version") limits our flexibility and is generally error prone. It creates an implicit contract with the backend that we should return such message if no new version is found.

backend-status/src/dbus/intf_impl.rs
AlexKaravaev: oh no, left after debugging, nice catch ty!

backend-status/src/dbus/intf_impl.rs
AlexKaravaev: this is clippy :(

update-agent/src/claim.rs
AlexKaravaev: but i am a bit afraid that if we get not_found without any msg(e.g backend broken and we get just 404) - we won't notice it then

Should i just check that any msg is returned?

backend-status/src/backend/types.rs
```diff
@@ -61,6 +62,7 @@ pub struct UpdateProgressV2 {
     pub install_progress: u64,
     pub total_progress: u64,
     pub error: Option<String>,
+    pub state: UpdateAgentState,
```
paulquinn00, line 175: Let's make UpdateAgentState a new top-level object in Status so we can deprecate UpdateProgressV2.

Noxcheborz (approved):

update-agent/src/claim.rs
AlexKaravaev: now i just check if it's not empty, but not sure it's the best

backend-status/src/dbus/intf_impl.rs
```diff
@@ -63,6 +66,42 @@ impl BackendStatusT for BackendStatusImpl {
         trace_ctx.apply(&span);
         let _guard = span.enter();
 
+        // Check if this is a critical state that should force immediate send
```
paulquinn00: I think there is cleaner way to do this. Every time a status is updated it notifies the main loop which runs `send_current_status`. That method calls 
```
    fn get_available_status(&self) -> Option<CurrentStatus> {
        if let Ok(mut current_status) = self.current_status.lock() {
            if self.last_update.elapsed() >= self.update_interval {
                return current_status.take();
            }
            // too soon to send again
        }
        None
    }
```
which is enforcing the periodicity.

I suggest putting in there a check to see if a reboot status is present and return the status to send immediate, rather than waiting the interval. This will cause an immediate send.

AlexKaravaev: thanks agreed! fixed in [0642347](https://github.com/worldcoin/orb-software/pull/571/commits/064234792e5119e756459bedc88663f5970280a6)

backend-status/src/update_progress.rs
```diff
@@ -18,132 +26,296 @@ pub enum UpdateProgressErr {
...
+            let (
+                download_progress,
+                processed_progress,
+                install_progress,
+                total_progress,
+            ) = match overall_state {
+                UpdateAgentState::Downloading => {
+                    (progress_value as u64, 0, 0, (progress_value / 3) as u64) // 1/3 of total workflow
+                }
+                UpdateAgentState::Fetched => (100, 0, 0, 66), // Downloading is still majority of time for the user
```
Michael-Preibisch, line 164: This should be 33, not 66. The Fetched state means "download complete" which is 1/3 of the total update process.

Currently causes the progress bar to jump from 33% to 66% instantly when downloads finish.

AlexKaravaev: > This should be 33, not 66. The Fetched state means "download complete" which is 1/3 of the total update process.

I was thinking for a bit on what to do here. Download **is** 1/3 of the whole process, **but** in fact for the user download is majority of the whole update process time due to the slow connection speed. So I think that if we hardcode it for 66(for now - we should improve it better) it would reflect a bit better for the actual user the overall installation progress

paulquinn00 (changes requested):

paulquinn00 (approved):

vmenge (approved):
lgtm

PR #572, opened 2025-07-21
feat(mcu-util): power cycle heat camera power & wifi/ble module

fouge:
wifi power can only be switched with a debug firmware

### DIFF
--- a/mcu-util/src/main.rs
+++ b/mcu-util/src/main.rs
@@ -59,9 +59,6 @@ enum SubCommand {
     /// Control UI
     #[clap(subcommand)]
     Ui(UiOpts),
-    /// Control secure element
-    #[clap(subcommand)]
-    SecureElement(SecureElement),
     /// Prints hardware revision from main MCU in machine-readable form
     #[clap(action)]
     HardwareRevision {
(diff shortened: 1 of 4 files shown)


mcu-util/src/main.rs
```diff
@@ -318,10 +323,16 @@ async fn execute(args: Args) -> Result<()> {
...
-        SubCommand::SecureElement(opts) => match opts {
-            SecureElement::PowerCycle => {
+        SubCommand::PowerCycle(opts) => match opts {
+            PowerCycleComponent::SecureElement => {
                 orb.sec_board_mut().power_cycle_secure_element().await?
             }
+            PowerCycleComponent::HeatCamera => {
+                orb.main_board_mut().heat_camera_power_cycle().await?
+            }
+            PowerCycleComponent::Wifi => {
```
AlexKaravaev, line 333: Should we give this much power 🤔

fouge: 🤔 
what do you mean?

fouge: should it be included into the binary you mean?

fouge: only the dev firmware supports the command
(but i tried and it's kind of breaking the whole network stack, even after reboot of the wifi module, and with an IP assigned, i wasn't able to reconnect locally)

AlexKaravaev: ah that's fine then, if the actual command doesn't work on prod lgtm

PR #573, opened 2025-07-21
feat(hil): expose nfs

TheButlah:
This PR does several things:
* adds the necessary packages to the `tegra-bash` FHSEnv to enable the use of the boot-from-ram bootloader0 scripts.
* Configures a known name for the network interface of the orb's RCM ethernet, based on VID and PID, and names it `orbeth0`. This allows us to replicate our configuration across many orbs and NUCs.
* Configures a networkmanager connection for `orbeth0`, in particular setting it as a "shared" connection, which will cause the orb to run a DHCP server to assign ip addresses to the orb - necessary for NFS.
* Configures the NUC to run a NFS server. Anything under `/srv` will be exposed to the orb over NFS, and only on the local ipv4 10.42.0.0/24 subnet.
* Configures port forwarding for the NFS server.

### DIFF
--- a/nix/machines/hil-common.nix
+++ b/nix/machines/hil-common.nix
@@ -19,6 +19,27 @@ in
 
   # Enable networking
   networking.networkmanager.enable = true;
+  networking.networkmanager.ensureProfiles.profiles = {
+    "Orb RCM Ethernet" = {
+      connection = {
+        autoconnect-priority = "-999";
+        id = "Orb RCM Ethernet";
+        interface-name = "enp0s20f0u1"; # TODO: hard coding?
+        type = "ethernet";
+        uuid = "eb829d4c-5314-38d1-98fc-a31b4be8892f"; # TODO make this not hard coded?
+      };
+      ethernet = { };
+      ipv4 = {
+        method = "shared"; # sets up DHCP server and shares internet
+      };
+      ipv6 = {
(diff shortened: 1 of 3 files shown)


nix/machines/hil-common.nix
```diff
@@ -19,6 +19,36 @@ in
...
+    };
+  };
+  # Give the jetson USB ethernet a known name
+  services.udev.extraRules = ''
+    ACTION=="add", \
+    SUBSYSTEM=="net", \
+    SUBSYSTEMS=="usb", \
+    ATTRS{idVendor}=="0955", \
+    ATTRS{idProduct}=="7035", \
+    NAME="orbeth0"
```
alekseifedotov, line 48: what if there is two orbs connected to the same NUC?

sfikastheo (approved):

nix/machines/hil-common.nix
TheButlah: then it wont work. We will have to cross that bridge when we get there. 1 orb working better than 0 orbs working.

TheButlah: is there a good way to solve it for the multiple orbs case? I am down to add it if so

TheButlah: https://github.com/worldcoin/orb-software/pull/576/files

PR #574, opened 2025-07-21
fix: Fix dbus env setting in worldcoin-jobs-agent service file

matthewpaul:
Also corrects the relay namespace that the orb-jobs-agent uses to communicate with the jobs server.

### DIFF
--- a/jobs-agent/debian/worldcoin-jobs-agent.service
+++ b/jobs-agent/debian/worldcoin-jobs-agent.service
@@ -11,8 +11,6 @@ Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/worldcoin_bus_socket
 Environment=RUST_BACKTRACE=1
 ExecStart=/usr/local/bin/orb-jobs-agent
 SyslogIdentifier=worldcoin-jobs-agent
-User=worldcoin
-Group=worldcoin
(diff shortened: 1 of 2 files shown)


jobs-agent/debian/worldcoin-jobs-agent.service
```diff
@@ -7,7 +7,9 @@ Requires=network.target
 Environment="HOME=/home/worldcoin"
 WorkingDirectory=/home/worldcoin
 Type=exec
-ExecStart=/bin/bash -c 'exec env ORB_ID="$(/usr/local/bin/orb-id)" /usr/local/bin/orb-jobs-agent'
+Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/worldcoin_bus_socket
+Environment=RUST_BACKTRACE=1
+ExecStart=/usr/local/bin/orb-jobs-agent
 SyslogIdentifier=worldcoin-jobs-agent
 User=worldcoin
 Group=worldcoin
```
sfikastheo: Can we also remove these two ?.
The service requires privileged environment. It should be run as root.

sfikastheo (changes requested):

sfikastheo (approved):

PR #579, opened 2025-07-23
refactor(orb-jobs-agent): reduce boilerplate, add tests, ctx 

vmenge:
refactors current jobs-agent impl.
main goal is not improvements in current _command_ implementations, but instead a refactoring of the framework around it, with minor fixes and a goal of reaching feature parity with old impl. 

## handler api / job famework
- commands now have a centralized routing place when building the `JobHandler` in `program.rs`, where all commands can be seen at a glance together with their parallel / sequential limit configuration
- handlers can now be async functions or closures
- argument parsing logic introduced
- extracted common calls from all handlers into the routing logic to reduce boilerplate needed when creating handlers such as:
  - checks for cancellation
  - marking job completion
  - returning `JobExecutionUpdate`
  - logging errors and successes
- introduced a `Ctx` for every handler to parameterize dependencies for testing, but also introduce helper methods to generate `JobExecutionUpdate` and to send progress reports for long living commands

## new commands
- extra logging command to log N lines from a service

## fixes
- reboot command now uses a store that can be injected on startup (`/mnt/update` for pearl, `/mnt/scratch` for diamond)

## tests
- addition of a fixture to quickly spin up a `fleet-cmdr` double for testing, or spin up new instances of `orb-jobs-agent`
- integration tests for most commands and the parallel / sequential job logic


jobs-agent/src/handlers/read_file.rs
```diff
@@ -0,0 +1,14 @@
...
+use color_eyre::{eyre::ContextCompat, Result};
+use orb_relay_messages::jobs::v1::JobExecutionUpdate;
+use tracing::info;
+
+pub async fn handler(ctx: Ctx) -> Result<JobExecutionUpdate> {
+    let file_path = ctx.args().first().wrap_err("no file path argument given")?;
+    info!("Reading file: {} for job {}", file_path, ctx.execution_id());
+
+    // Execute the file reading logic
+    let result = tokio::fs::read(file_path).await?;
```
vmenge, line 12: /fp go away

jobs-agent/src/handlers/logs.rs
```diff
@@ -0,0 +1,119 @@
...
+        LogAction::Tail { service } => {
+            let now = Instant::now();
+            let max_duration = Duration::from_secs(5 * 60);
+
+            // send initial progress
+            let _ = ctx.progress().send().await;
+            let mut child_proc = ctx
+                .deps()
+                .shell
+                .exec(&["journalctl", "-f", "-u", &service])
```
alekseifedotov, line 62: It is not the best way to read logs, journalctl will miss extra tags logged by the service. For example tracing framework uses them to provide extra info.

vmenge: thanks for the heads up! right now just trying to keep it similar to previous logic and migrate to new refactor and tests, then i'll try any improvements

jobs-agent/src/handlers/reboot.rs
```diff
@@ -0,0 +1,87 @@
...
+    Pending(String),
+    /// We are free to perform a reboot
+    Free,
+}
+
+impl RebootStatus {
+    const FILENAME: &str = "reboot.lock";
+
+    async fn from_lockfile(store: impl AsRef<Path>) -> Result<Self> {
+        match fs::read_to_string(store.as_ref().join(Self::FILENAME)).await {
```
vmenge, line 65: /fp useless bot

Cargo.toml
```diff
@@ -69,14 +69,15 @@ members = [
 edition = "2024"
 license = "MIT OR (Apache-2.0 WITH LLVM-exception)"
 repository = "https://github.com/worldcoin/orb-software"
-rust-version = "1.87.0" # See rust-toolchain.toml
+rust-version = "1.88.0" # See rust-toolchain.toml
```
TheButlah, line 72: For the future: Changes like this should be done in their own pr, to keep diffs small and minimize merge conflicts

jobs-agent/Cargo.toml
```diff
@@ -56,11 +71,18 @@ rev = "ee0d9f4877e7d4a7dcd99df84649dc0a89235b2e"
...
 tokio = { workspace = true, features = ["sync", "test-util"] }
+async-tempfile = { workspace = true }
+testcontainers = "0.23.2"
 
 [package.metadata.deb]
 maintainer-scripts = "debian/"
 assets = [["target/release/orb-jobs-agent", "/usr/local/bin/", "755"]]
 systemd-units = [{ unit-name = "worldcoin-jobs-agent" }]
+
+[package.metadata.orb]
```
TheButlah: What makes it unsupported?

TheButlah: I've confirmed it compiles on macos, so I will fix this part

PR #580, opened 2025-07-24
Always powercycle thermal camera on pairing

AlexKaravaev:


AlexKaravaev:
@fouge should i check for pearl vs. diamond here before power cycling it, or firmware will just ignore the message on pearl without error and we will be fine?

fouge:
@AlexKaravaev 
> without error
you will have an error returned

just tested:
```
$ ./orb-mcu-util power-cycle heat-camera
2025-07-24T07:31:48.654744Z ERROR orb_mcu_util: "error power cycling heat camera (2v8): ack OperationNotSupported"
$ echo $?
255
```

thermal-cam-ctrl/src/pairing.rs
```diff
@@ -102,6 +107,28 @@ impl Pair {
 ///////////////////////////
 // ---- Helper Code ---- //
 ///////////////////////////
+fn power_cycle_heat_camera() -> Result<()> {
+    let orb_os_release =
+        OrbOsRelease::read_blocking().wrap_err("Failed to read /etc/os-release")?;
```
TheButlah: instead of retrieving this from env vars or files on the system (which makes it hard to tell at a glance what this binary depends on os-wise), please either:
* pass this configuration in explicitly, from main. So your env var detection would be in main, not here.
* add a required CLI arg, and let orb-os use a systemd override to set that cli arg on diamond.

thermal-cam-ctrl/src/pairing.rs
```diff
@@ -102,6 +107,28 @@ impl Pair {
...
+            let status = Command::new("orb-mcu-util")
+                .args(["power-cycle", "heat-camera"])
+                .status()
+                .wrap_err("Failed to execute orb-mcu-util power-cycle heat-camera")?;
+            if !status.success() {
+                return Err(eyre!(
+                    "orb-mcu-util power-cycle heat-camera exited with non-zero status: {status}"));
+            }
+        }
+        OrbOsPlatform::Pearl => {}
```
TheButlah, line 124: Is it impossible to power cycle the camera on pearl? If its possible, we should power cycle on pearl too.

AlexKaravaev: it is impossible, at least for now...

thermal-cam-ctrl/src/pairing.rs
AlexKaravaev: hm, but I am not convinced why it's better? orb-info crate is our default place for storing everything that requires asserting on orb platform state, we also assume that it always correctly populated, because we depend on it in more critical places like update-verifier

TheButlah: regardless you should pass in the configuration from main explicitly, rather than implicitly grab it from the environment.

Environment specific configuration should be treated as configuration that is passed as arguments into the things that needs it. regardless of if using orb-info or not.

AlexKaravaev: got it, agreed and fixed

PR #586, opened 2025-07-26
feat: migrate to new update endpoint

AlexKaravaev:


update-agent/src/claim.rs
```diff
@@ -147,6 +152,16 @@ fn from_remote(
     } else {
         let resp_txt = resp.text().map_err(Error::ResponseAsText)?;
         debug!("server sent raw claim: {resp_txt}");
+
+        if let Ok(response) = serde_json::from_str::<serde_json::Value>(&resp_txt) {
```
TheButlah, line 156: any particular reason we are using serde_json::Value instead of an actual struct?

AlexKaravaev: I think that's fine since we are just doing early return? No need to deserialize here

Michael-Preibisch (approved):

update-agent/src/claim.rs
TheButlah: This *does* deserialize though, just not into a real type

PR #589, opened 2025-07-30
feat(blob-transfer): wip

pophilpo:
wip for hackathon project

### DIFF
--- a/orb-blob/blob-transfer/src/lib.rs
+++ b/orb-blob/blob-transfer/src/lib.rs
@@ -1,20 +1,25 @@
 use std::{path::Path, sync::Arc};
 
 use color_eyre::Result;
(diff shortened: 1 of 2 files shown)


TheButlah (approved):

orb-blob/blob-transfer/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Ok(Self {
+            endpoint,
+            store,
+            _router: router,
+        })
+    }
+
+    pub async fn import(&self, path: &Path) -> Result<BlobTicket> {
+        let abs_path = std::fs::canonicalize(path)?;
```
TheButlah, line 38: blocks in async @pophilpo

PR #601, opened 2025-08-01
fix(blobs): create inner task for listening

TheButlah:
all clients now subscribe and manage listening to the topic with a task.


pophilpo (approved):
based :1st_place_medal:

pophilpo (approved):

PR #617, opened 2025-08-07
feat(hil): switch nixos.raw.zst to liveusb.raw.gz

TheButlah:
For macos users without brew/zstd installed.


TheButlah:
confirmed that it works on my system76

PR #624, opened 2025-08-12
fix(jobs-agent): race condition whenrequesting a new job

vmenge:
## fixes
- race condition if a job completes a job, but ends up requesting a new job before the server has time to remove it from its queue

## tests
- test for the above race  condition
- test we return FailedUnsupported for unknown jobs

## docs
- updated readme

### DIFF
--- a/jobs-agent/src/job_system/handler.rs
+++ b/jobs-agent/src/job_system/handler.rs
@@ -341,6 +341,8 @@ impl JobHandler {
         let job_client = self.job_client.clone();
         let job_clone = job.clone();
         let update = ctx.status(JobExecutionStatus::InProgress);
+        let job_registry = self.job_registry.clone();
+
         tokio::spawn(async move {
             if ctx.is_cancelled() {
                 let update = ctx.failure().stdout("Job was cancelled");
@@ -351,6 +353,16 @@ impl JobHandler {
 
                 let _ = completion_tx
                     .send(JobCompletion::cancelled(ctx.execution_id().to_owned()));
+
+                return;
+            }
+
+            if job_registry.is_job_completed(ctx.execution_id()).await {
+                info!(
+                    "trying to execute a job that was recently completed. ignored. {} {}",
+                    ctx.cmd(),
+                    ctx.execution_id()
+                );
 
                 return;
             }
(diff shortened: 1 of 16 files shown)


jobs-agent/src/job_system/handler.rs
```diff
@@ -452,6 +464,11 @@ impl JobHandler {
                     // Unregister job
                     job_registry_clone.unregister_job(&job_execution_id).await;
 
+                    if completion.status == JobExecutionStatus::InProgress {
```
matthewpaul, line 467: I'm not sure I understand the purpose of this conditional. Can you explain?

vmenge: if a job returns in progress from the job, its not really completed -- i.e., the reboot job. it was requesting for another job even though its not actually finished with the job and was just waiting for a reboot (we can  both finish a job with a `JobExecutionStatus::InProgress`, as well as stream progress updates with `ctx.progress()` like in the log handling)

test-utils/src/async_bag.rs
```diff
@@ -1,9 +1,25 @@
 use std::{ops::Deref, sync::Arc};
 use tokio::sync::{Mutex, MutexGuard};
 
-#[derive(Clone, Debug)]
+/// Wrapper for `Arc<Mutex<T>>` used to make testing easier.
+#[derive(Debug)]
 pub struct AsyncBag<T>(Arc<Mutex<T>>);
```
TheButlah, line 6: whats the point of this struct?

vmenge: avoiding `Arc::new(Mutex::new())` plus some utility methods

TheButlah: imo it is harder to understand what an `AsyncBag<T>` is than an `Arc<Mutex<T>>`. Also, theres nothing gained by added type safety here, so a type alias would be preferable

TheButlah: side note: The tokio mutex is an antipattern. Its extremely rare that it should be used. typically you want a std mutex, even in async code (assuming you don't hold on to locks a long time, but if you held onto locks a long time, you probably should have used a channel)

PR #634, opened 2025-08-15
feat(hil): scaffold nfsboot subcommand

TheButlah:


AlexKaravaev (approved):
👍 should we add some docs to mdbook, or too early?

TheButlah:
Too early

PR #635, opened 2025-08-15
fix(orb-registration): Diamond Default Channel

chrisgalanis:
This pull request updates the logic for selecting the channel name in the `orb-registration.py` script when using the `prod` backend. The change ensures that the channel is set to `diamond-tier-ga` specifically when both the platform is `diamond` and the channel is `general`.


pophilpo (approved):

sfikastheo:
channels are defined by `CaC`, i believe that this is just tech-dept.

AlexKaravaev (approved):

PR #639, opened 2025-08-15
fix: storage discovery

alekseifedotov:
### Current state
Depending on the orb model, storage device could be:
- Pearl - `/dev/mmcblk0`
- Diamond with SD Express card: `/dev/nvme0n1`
- Diamond with SD card `/dev/mmcblk1`

In the past that was expressed in the claim, by specifying device type 'nvme'(for `/dev/nvme0n1`) or 'emmc' (for `/dev/mmcblk0`). 

### Problem
We need to add support for Diamond with plain SD card (`/dev/mmcblk1`). We have an option to add yet another device type, but that will make manifest generation more complicated and will split the fleet even further. 

### Proposed solution
Instead of adding more device types. I propose to have only one device type SSD (solid state drive).  And detect the storage device used by checking where the `/usr/persistent/` is mounted.

### DIFF
--- a/update-agent/core/src/components.rs
+++ b/update-agent/core/src/components.rs
@@ -67,10 +67,7 @@ impl<'de> Deserialize<'de> for Device {
 
 impl Device {
     fn to_path(&self) -> PathBuf {
-        match self {
-            Self::Ssd => PathBuf::from(&self.to_string()),
-            Self::Qspi => PathBuf::from(&self.to_string()),
-        }
+        PathBuf::from(&self.to_string())
     }
 }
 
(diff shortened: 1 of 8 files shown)


update-agent/core/src/components.rs
```diff
@@ -20,37 +25,109 @@ pub enum Location {
...
+/// echo "/dev/$device_name"
+fn find_block_device_by_mountpoint(
+    mountpoint: &std::path::Path,
+) -> std::io::Result<PathBuf> {
+    // 'stat' the mountpoint. (see man 2 stat)
+    let metadata = std::fs::metadata(mountpoint)?;
+
+    // Get major & minor of the underlying device
+    let dev = metadata.dev();
+    let major = dev >> 8;
```
TheButlah: use https://docs.rs/rustix/latest/rustix/fs/fn.major.html and https://docs.rs/rustix/latest/rustix/fs/fn.minor.html.

Yes, I know its a dependency for only 1 thing that could be a bitwise operation. But in my defense, the linux docs say we should use the functions too. https://www.man7.org/linux/man-pages/man3/major.3.html

TheButlah (approved):

PR #641, opened 2025-08-16
fix: update agent mock tests

AlexKaravaev:
fix functional tests of update-agent


alekseifedotov:
fixed in https://github.com/worldcoin/orb-software/pull/639

PR #647, opened 2025-08-19
feat(backend-status): lte_info

vmenge:
adds `LteInfo` struct and corresponding dbus method to receive the info

### DIFF
--- a/backend-status/dbus/src/types.rs
+++ b/backend-status/dbus/src/types.rs
@@ -64,19 +64,21 @@ pub struct NetIntf {
     pub rx_errors: u64,
 }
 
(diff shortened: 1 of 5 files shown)


backend-status/dbus/src/types.rs
pophilpo, line 1: ```/// Signal-to-Noise Ratio) — how "clean" the signal is.```
Can you remove the ")" here ?

vmenge: no u (yes)

vmenge: done :saluting_face:

PR #648, opened 2025-08-20
chore(jobs-agent): disable reboot command

vmenge:
disabling the reboot command temporarily due to:
- nvidia retry counter decreasing on every reboot (as opposed to full power cycle)
- `orb-mcu-util reboot main` has a bug where orb isn't powered back on, so we can't use it instead yet

### DIFF
--- a/jobs-agent/src/program.rs
+++ b/jobs-agent/src/program.rs
@@ -1,5 +1,5 @@
 use crate::{
-    handlers::{check_my_orb, logs, mcu, orb_details, read_file, read_gimbal, reboot},
+    handlers::{check_my_orb, logs, mcu, orb_details, read_file, read_gimbal},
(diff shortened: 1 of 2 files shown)


pophilpo (approved):

jobs-agent/src/program.rs
```diff
@@ -36,7 +36,8 @@ pub async fn run(deps: Deps) -> Result<()> {
         .parallel("read_gimbal", read_gimbal::handler)
         .parallel("mcu", mcu::handler)
         .parallel_max("logs", 3, logs::handler)
-        .sequential("reboot", reboot::handler)
+        // .sequential("reboot", reboot::handler) ignored for now, mcu reboot is broken, and
```
fouge, line 39: is mcu reboot broken?

PR #649, opened 2025-08-20
fix: abort the build if building for non-linux

alekseifedotov:
Fixing AI slop.

There are 2 problems fixed here:
- building for non-linux is actually a fatal error, not a warning
- to report an error it needs to print 'cargo::error' (see two semicolons), but LLM wrote it with 1 error.

### DIFF
--- a/update-agent-loader/build.rs
+++ b/update-agent-loader/build.rs
@@ -6,7 +6,7 @@ use std::fs;
 fn main() {
     // This package is Linux-only, skip build on other platforms
     if !cfg!(target_os = "linux") {
-        println!("cargo:warning=update-agent-loader is only supported on Linux, skipping build");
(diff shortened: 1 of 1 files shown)


update-agent-loader/build.rs
```diff
@@ -6,7 +6,7 @@ use std::fs;
 fn main() {
     // This package is Linux-only, skip build on other platforms
     if !cfg!(target_os = "linux") {
-        println!("cargo:warning=update-agent-loader is only supported on Linux, skipping build");
+        println!("cargo::error=update-agent-loader is only supported on Linux, skipping build");
```
TheButlah, line 9: Imo this build scriot should be deleted wholesale, just let it fail to build on linux

TheButlah:
Closing in favor of #650

PR #650, opened 2025-08-20
refactor(update-agent-loader): pass configuration in explicitly

TheButlah:


update-agent-loader/src/memfile.rs
```diff
@@ -75,11 +71,13 @@ impl From<io::Error> for MemFileError {
 /// An in-memory file created with memfd_create
 pub struct MemFile<S: MemFileState> {
     fd: OwnedFd,
-    _marker: PhantomData<S>,
+    marker: S,
```
alekseifedotov: why PhantomData is gone? Now `marker` is not a zero sized type.

update-agent-loader/src/memfile.rs
```diff
@@ -156,19 +154,9 @@ impl<S: MemFileState> MemFile<S> {
...
-        .decode(env!("PUBLIC_KEY_BASE64"))
-        .unwrap()
-        .try_into()
-        .unwrap();
-}
-
 impl MemFile<Unverified> {
     /// Create a new empty memory file with close-on-exec flag
-    pub fn create() -> Result<Self, MemFileError> {
+    pub fn create(pubkey: VerifyingKey) -> Result<Self, MemFileError> {
```
alekseifedotov: Why do you pass `pubkey` when creating an unverified memfile?

The pubkey should be passed to `verify_signature` and not here.

alekseifedotov (changes requested):

PR #655, opened 2025-08-21
feat!: Add new QR version for new session flow

danielle-tfh:
### Notes
We are planning a new roll-out of the session validation flow: https://www.notion.so/worldcoin/Proposal-Eliminate-Session-Centric-Verification-Flow-2418614bdf8c802893d9c3fe43ebba81.

This PR prepares a new version of qr codes where the orb knows the version of the QR code

* A description of the new flow is written in lib
* relies on the new orb relay message types specified in https://github.com/worldcoin/orb-relay-messages/pull/71


qr-link/src/user_data.rs
```diff
@@ -1,12 +1,13 @@
 use blake3::Hasher;
 use serde::{Deserialize, Serialize};
 
-const PCP_VERSION_DEFAULT: u16 = 2;
+pub const PCP_VERSION_DEFAULT: u16 = 2;
```
andronat: why do we need this public? is it for orb relay messages?

andronat (changes requested):
Sorry i might be too tired but i think this won't work. if apps upgrade to thid version the hash will be calculated differently between app and orb. also if we upgrade the orbs first, we will have the same issue. both chancing fields have issues. i think the only way to roll this out is either by doing 2 hash functions or by 2 releases. let's discuss tomorrow. again i might be completely wrong and my brain hallucinate..

qr-link/src/app_authenticated_data.rs
```diff
@@ -0,0 +1,59 @@
...
+            pcp_version,
+            os_version,
+            os,
+        } = self;
+        hasher.update(identity_commitment.as_bytes());
+        hasher.update(self_custody_public_key.as_bytes());
+        hasher.update(os_version.as_bytes());
+        hasher.update(os.as_bytes());
+        if *pcp_version != PCP_VERSION_DEFAULT {
+            hasher.update(&pcp_version.to_ne_bytes());
```
andronat: Hmm actually this is dangerous no? `ne` is for native endianness and if different platforms have different endianness this will break... We are actually lucky that `arm` and `x86` are little endian by default. Could we do it `_le_`?

qr-link/src/decode.rs
```diff
@@ -17,6 +17,8 @@ pub enum DecodeError {
 }
 
 /// Parses `session_id` and `user_data_hash` from a QR-code string.
+/// This decode version does not support v4 since no v4 should be used with this method
+/// All orbs should be updated to support v4 since it requires specific logic
 pub fn decode_qr(qr: &str) -> Result<(Uuid, Vec<u8>), DecodeError> {
```
andronat: I think we should remove this or break the current api. in any case the code doesn't need to be backwards compatible. Only the QR handling needs to. The plan will be we update all the orbs first to have the new v3+v4 decoders everywhere and then we update the apps to generate v4 only QRs.

qr-link/src/encode.rs
```diff
@@ -15,3 +16,18 @@ pub fn encode_qr(session_id: &Uuid, user_data_hash: impl AsRef<[u8]>) -> String
     BASE64_NOPAD.encode_append(&payload, &mut qr);
     qr
 }
+
+/// Generates a QR-code (V4) string from `orb relay id` and `app_authenticated_data`
+pub fn encode_static_qr(
```
andronat, line 7: Similar to my comments above, we should remove the old `encode_qr` and just update it with the following code. We no longer need to generate v3s if we move to v4.

qr-link/src/lib.rs
```diff
@@ -75,18 +76,74 @@
 //! // from the backend.
 //! let success = user_data.verify(user_data_hash);
 //! ```
+//!
+//! # New flow
```
andronat: Similar here. Just remove the old examples and make this the canonical one.

andronat (approved):

qr-link/src/app_authenticated_data.rs
danielle-tfh: Nice find! Let me update it

qr-link/src/encode.rs
danielle-tfh: I initially kept it for  unit tests for ensuring backwards compatibility with decode.

But if you feel that its not needed - then let me remove

PR #680, opened 2025-08-22
feat(backend-status): FM-507-backend-status-signup-state

paulquinn00:
Added collector to watch for signals from `org.worldcoin.OrbCore1.Signup`
- reporting state of signup readiness from OrbCore

This helps the backend determine if Orb Core is waiting for an Operator QR code.


orb-backend-status/src/dbus/proxies.rs
vmenge, line 1: hey sorry, i looked and couldn't find usage of this anywhere in this repo :thinking: 
is this being used somewhere else?

orb-backend-status/src/collectors/core_signups.rs
```diff
@@ -0,0 +1,240 @@
...
+            let serialized =
+                serde_json::to_string(&state).expect("Failed to serialize");
+            let deserialized: SignupState =
+                serde_json::from_str(&serialized).expect("Failed to deserialize");
+            assert_eq!(state, deserialized);
+        }
+    }
+
+    #[tokio::test]
+    async fn test_signup_watcher_initialization() -> Result<()> {
```
vmenge, line 204: could you add one more test for reacting to the signals? :pray:

vmenge (changes requested):
lgtm!
just want to understand where the proxy (i left a comment) is being used, and if you could add one more test for the watcher that would be awesome :pray:

PR #683, opened 2025-08-24
fix(backend-status): remove oververbose logging for update agent messages

AlexKaravaev:
remove oververbose logging


TheButlah (approved):
consider switching it to either trace or debug.

PR #689, opened 2025-08-28
feat(connd): handle cellular connection on start up, including reconnection

vmenge:
## about
`connd` will establish and maintain cellular connectivity on all diamond orbs with a modem and eSIM (except in service image)

## new 
- `connd` will establish a connection through cellular on boot (if modem is available) with a logic similar to `orb-wwan-init`
- `connd` will periodically check for connection status, and in case of connection loss will drop bearers with `simple_disconnect` and attempt the whole process again

## todo
- [x] add config option so connd does not connect automatically on service images
- [x] test: connecting to wifi through a QR code after establishing cellular connection
- [x] test: connecting to both wifi and cellular, then losing connection and maintaining connectivity through cellular without issues

above tests depends on these changes in orb-core:
https://github.com/worldcoin/priv-orb-core/pull/1713


sfikastheo (approved):
Marking: this (in theory) is an intermediate step. NetworkManager may be managing most of this in the future TM

PR #695, opened 2025-09-01
fix: Fix update agent progress reporting

AlexKaravaev:
Now tested e2e with stage, it reports it correctly

### DIFF
--- a/orb-backend-status/src/collectors/update_progress.rs
+++ b/orb-backend-status/src/collectors/update_progress.rs
@@ -237,8 +237,8 @@ impl UpdateProgressWatcher {
             None
         };
 
-Ok((overall_progress.is_some() || overall_state.is_some())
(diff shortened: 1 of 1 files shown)


orb-backend-status/src/collectors/update_progress.rs
```diff
@@ -221,19 +224,24 @@ impl UpdateProgressWatcher {
...
-            return Ok(None);
+            None
         };
 
-        Ok(Some((overall_progress, overall_state)))
+        if overall_progress.is_some() || overall_state.is_some() {
+            Ok(Some((overall_progress, overall_state)))
+        } else {
+            Ok(None)
+        }
```
sfikastheo:
```suggestion
Ok((overall_progress.is_some() || overall_state.is_some())
    .then_some((overall_progress, overall_state)))
```

PR #702, opened 2025-09-05
fix(deny): unmaintained crates, and allow NCSA license

fouge:
to use adler2 instead of the unmaintained adler crate
update `image` dep (and remove from mcu-util, unused)

### DIFF
--- a/ui/cone/Cargo.toml
+++ b/ui/cone/Cargo.toml
@@ -15,7 +15,7 @@ embedded-graphics = "0.8.1"
 ftdi-embedded-hal.workspace = true
 futures.workspace = true
 gc9a01-rs = "0.2.1"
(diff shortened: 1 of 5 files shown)


Qbicz (approved):

ui/cone/Cargo.toml
```diff
@@ -15,7 +15,7 @@ embedded-graphics = "0.8.1"
 ftdi-embedded-hal.workspace = true
 futures.workspace = true
 gc9a01-rs = "0.2.1"
-image = { version = "=0.25.1", features = [] }
+image = { version = "=0.25.8", features = [] }
```
TheButlah, line 18: typically you should not use `=` dependencies unless theres a good reason

PR #703, opened 2025-09-05
feat: Add support for --json flag in check-my-orb and new wifi_ip handler

matthewpaul:
This pull request adds a new handler for retrieving the WiFi IP address and improves the flexibility of the `check_my_orb` command. It also updates the systemd service dependencies to ensure proper startup order. The most important changes are grouped below:

### New WiFi IP Handler

* Added the `wifi_ip` handler to provide the IP address of the currently connected WiFi interface, including detection logic and output formatting. (`orb-jobs-agent/src/handlers/wifi_ip.rs`)
* Registered the new `wifi_ip` handler in the module exports and the job system, making it available for use. (`orb-jobs-agent/src/handlers/mod.rs`, `orb-jobs-agent/src/program.rs`) [[1]](diffhunk://#diff-828c40d1bb4e8a2bc1a35351f4a0549510dc8c0b004035a96277f1a83096a409R8) [[2]](diffhunk://#diff-6ef7e04362a7d78e147b3e7ed0702028b79c0fe368c0284d1ba85f96b0664f28L2-R2) [[3]](diffhunk://#diff-6ef7e04362a7d78e147b3e7ed0702028b79c0fe368c0284d1ba85f96b0664f28R38)

### Improvements to Command Handling

* Enhanced the `check_my_orb` handler to support an optional `--json` flag, allowing output in either default or JSON format based on user input. (`orb-jobs-agent/src/handlers/check_my_orb.rs`)

### Service Dependency Update

* Changed the systemd service file to require and start after `worldcoin-attest.service`, ensuring the jobs agent starts in the correct order relative to other services. (`orb-jobs-agent/debian/worldcoin-jobs-agent.service`)

### DIFF
--- a/orb-jobs-agent/debian/worldcoin-jobs-agent.service
+++ b/orb-jobs-agent/debian/worldcoin-jobs-agent.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Worldcoin Orb Jobs Agent
-After=network.target
-Requires=network.target
+Requires=worldcoin-attest.service
+After=worldcoin-attest.service
 
 [Service]
 Environment="HOME=/home/worldcoin"
--- a/orb-jobs-agent/src/handlers/wifi_ip.rs
+++ b/orb-jobs-agent/src/handlers/wifi_ip.rs
@@ -18,7 +18,7 @@ pub async fn handler(ctx: Ctx) -> Result<JobExecutionUpdate> {
         .wrap_err("failed to get output for route command")?;
 
     let route_info = String::from_utf8_lossy(&route_output.stdout);
-    
+
     // Extract the interface name from the route output
     let interface = route_info
         .split_whitespace()
@@ -32,13 +32,15 @@ pub async fn handler(ctx: Ctx) -> Result<JobExecutionUpdate> {
         .shell
         .exec(&["ip", "addr", "show", interface])
         .await
-        .wrap_err_with(|| format!("failed to get IP address for interface {}", interface))?
+        .wrap_err_with(|| {
+            format!("failed to get IP address for interface {interface}")
+        })?
         .wait_with_output()
         .await
         .wrap_err("failed to get output for ip addr command")?;
 
     let ip_info = String::from_utf8_lossy(&ip_output.stdout);
-    
+
     // Parse the IP address from the output
     let ip_address = ip_info
         .lines()
--- a/orb-jobs-agent/src/handlers/wifi_ip.rs
+++ b/orb-jobs-agent/src/handlers/wifi_ip.rs
@@ -1,0 +1,60 @@
+use crate::job_system::ctx::{Ctx, JobExecutionUpdateExt};
+use color_eyre::{eyre::Context, Result};
+use orb_relay_messages::jobs::v1::JobExecutionUpdate;
+
+/// command format: `wifi_ip`
+#[tracing::instrument]
+pub async fn handler(ctx: Ctx) -> Result<JobExecutionUpdate> {
+    // Try to get the IP address of the currently connected WiFi interface
+    // First, try to get the default route interface
+    let route_output = ctx
+        .deps()
+        .shell
+        .exec(&["ip", "route", "get", "8.8.8.8"])
+        .await
+        .wrap_err("failed to get default route")?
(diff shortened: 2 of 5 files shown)


orb-jobs-agent/debian/worldcoin-jobs-agent.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Worldcoin Orb Jobs Agent
-After=network.target
-Requires=network.target
+Requires=worldcoin-attest.service
+After=worldcoin-attest.service
```
AlexKaravaev, line 4: Would need @vmenge input here, but I think he designed jobs agent to always poll for token, so we don't need explicit dependency

orb-jobs-agent/src/handlers/wifi_ip.rs
```diff
@@ -0,0 +1,60 @@
...
+        .wrap_err("failed to get output for route command")?;
+
+    let route_info = String::from_utf8_lossy(&route_output.stdout);
+    
+    // Extract the interface name from the route output
+    let interface = route_info
+        .split_whitespace()
+        .skip_while(|&word| word != "dev")
+        .nth(1)
+        .unwrap_or("wlan0"); // fallback to common WiFi interface name
```
AlexKaravaev, line 27: Side note, we do need this command in jobs agent for sure? I think backend-status already provides this

orb-jobs-agent/src/handlers/wifi_ip.rs
```diff
@@ -0,0 +1,60 @@
...
+        .exec(&["ip", "route", "get", "8.8.8.8"])
+        .await
+        .wrap_err("failed to get default route")?
+        .wait_with_output()
+        .await
+        .wrap_err("failed to get output for route command")?;
+
+    let route_info = String::from_utf8_lossy(&route_output.stdout);
+    
+    // Extract the interface name from the route output
```
AlexKaravaev, line 22: nit: but ip supports --json flag - can be easier to parse, but really just a nit and fyi
```
(venv) (os-2.25.2-rc-diamond.35855972471d2930c39ef05d298c7ba91a429616.dev) worldcoin@id-a160b63b:~$ ip -j route get 8.8.8.8
[{"dst":"8.8.8.8","gateway":"10.108.0.1","dev":"wlan0","prefsrc":"10.108.8.174","flags":[],"uid":1000,"cache":[]}]
```

orb-jobs-agent/debian/worldcoin-jobs-agent.service
AlexKaravaev: before merging I would just need to hear @vmenge about this, but else is good! 🚀

orb-jobs-agent/src/handlers/wifi_ip.rs
matthewpaul: I am not certain, but it's proving useful for development with stage orbs

orb-jobs-agent/src/handlers/wifi_ip.rs
matthewpaul: Current output from the job looks like this:

{"interface":"wlan0","ip_address":"10.108.0.68","status":"connected"}

orb-jobs-agent/debian/worldcoin-jobs-agent.service
sfikastheo: `Requires` and `After` orb-attest does not mean that you have an attestation token.
You have to poll for it, but the dependency here is correct. 
Since the `job-agent` needs an attestation token it makes sense that the service gets started after the `attestation` service IMO.

sfikastheo: Additionally `network.target` means "little" for the `jobs-agent`. +1 for the change.

sfikastheo (approved):

orb-jobs-agent/debian/worldcoin-jobs-agent.service
vmenge: Nothing to add more than what @AlexKaravaev and @sfikastheo said, change looks good to me!

PR #709, opened 2025-09-09
fix(jobs-agent): reboot will now reboot but not by rebooting but instead by shutting down and power cycling beceuase reboot is a no no

vmenge:
## fix
- use newly added reboot command in `orb-mcu-util` (thanks @fouge!)
- orb will now power cycle as a reboot, being powered back on by the mcu
- this will avoid issues with nvidia's retry counter which went down with regular reboots

## todo
- [x] test on a real orb :D

reboooot


orb-jobs-agent/src/handlers/reboot.rs
```diff
@@ -41,8 +42,12 @@ pub async fn handler(ctx: Ctx) -> Result<JobExecutionUpdate> {
                 .await
                 .map_err(|e| eyre!("failed to send progress {e:?}"))?;
 
-            // TODO: switch to MCU reboot once that is fixed
-            ctx.deps().shell.exec(&["reboot"]).await?;
+            ctx.deps()
+                .shell
+                .exec(&["orb-mcu-util", "reboot", "orb", "--delay", "60"])
+                .await?;
```
fouge: should we still reboot the jetson even if mcu command fails?

vmenge: how would we do it though? we can't call `reboot` rn :thinking:

vmenge: why is the thinking emoji so aggressive here

PR #713, opened 2025-09-10
feat(s3-helpers): implement multipart upload

TheButlah:
Vibe coded with codex + gpt5. I have reviewed the code.

It did actually well, specifically because I told it to mimic the same styles used in my manually-implemented multipart_download from prior PRs. I think the approach of "manually implement it first, get it good, then have the AI clone the pattern" is a good way to use things. This is a different approach to how I usually see it used, which is "please go do this feature" (which typically results in pretty poor code).

Was pretty impressed that it oneshot it. I only did 1 minor nitpick tweak to switch the page_number to NonZeroU16. It probably helps that doing an s3 upload is an *extremely* common task.

### DIFF
--- a/s3-helpers/src/upload.rs
+++ b/s3-helpers/src/upload.rs
@@ -10,6 +10,7 @@ use color_eyre::{
     eyre::{bail, ensure, eyre, WrapErr as _},
     Result,
 };
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::{sync::Mutex, task::JoinSet, time::timeout};
 use tracing::warn;
 
@@ -45,9 +46,13 @@ pub(crate) async fn upload_multipart(
         "directories are not supported, make sure the s3 uri doesn't end in a slash"
     );
 
-    let metadata = tokio::fs::metadata(in_path)
+    let mut in_file = tokio::fs::File::open(in_path)
         .await
-        .wrap_err("failed to stat input file")?;
+        .wrap_err_with(|| format!("failed to open {in_path}"))?;
+    let metadata = in_file
+        .metadata()
+        .await
+        .wrap_err_with(|| format!("failed to stat {in_path}"))?;
     ensure!(metadata.is_file(), "input path must be a file");
     let total_bytes = metadata.len();
 
@@ -66,7 +71,9 @@ pub(crate) async fn upload_multipart(
 
     // For small files, use a single PutObject.
     if total_bytes <= PART_SIZE {
-        let bytes = tokio::fs::read(in_path)
+        let mut bytes = Vec::new();
+        in_file
+            .read_to_end(&mut bytes)
             .await
             .wrap_err("failed to read input file")?;
 
@@ -98,13 +105,11 @@ pub(crate) async fn upload_multipart(
         .ok_or_else(|| eyre!("upload id missing"))?
         .to_string();
 
-    let file = tokio::task::spawn_blocking({
-        let path = in_path.to_owned();
-        move || std::fs::File::open(path)
-    })
-    .await
-    .wrap_err("task panicked")??;
-    let file = Arc::new(file);
+    in_file
+        .flush()
+        .await
+        .wrap_err_with(|| format!("failed to flush {in_path}"))?;
+    let file = Arc::new(in_file.try_into_std().expect("infallible"));
 
     let ranges_iter = {
         let mut part_number: u16 = 1;


s3-helpers/src/upload.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+pub(crate) async fn upload_multipart(
+    client: &Client,
+    object: &S3Uri,
+    in_path: &Utf8Path,
+    existing_object_behavior: ExistingObjectBehavior,
+    mut progress: impl FnMut(UploadProgress),
+) -> Result<()> {
+    ensure!(
```
alekseifedotov, line 44: I wonder if that could be experssed in a type system. Like have different types for directory and files.

s3-helpers/src/upload.rs
```diff
@@ -0,0 +1,286 @@
...
+        !object.is_dir(),
+        "directories are not supported, make sure the s3 uri doesn't end in a slash"
+    );
+
+    let metadata = tokio::fs::metadata(in_path)
+        .await
+        .wrap_err("failed to stat input file")?;
+    ensure!(
+        metadata.is_file(),
+        "input path must be a file, not a directory"
```
alekseifedotov: It could also be a UNIX socket, or a device, so the message is not correct.

s3-helpers/src/upload.rs
```diff
@@ -0,0 +1,286 @@
...
+        .send()
+        .await
+        .wrap_err("failed to create multipart upload")?
+        .upload_id()
+        .ok_or_else(|| eyre!("upload id missing"))?
+        .to_string();
+
+    let file = tokio::task::spawn_blocking({
+        let path = in_path.to_owned();
+        move || std::fs::File::open(path)
```
alekseifedotov: why does it need a blocking task? can't it read the file using async primitives and create an async iterator of chunks?

alekseifedotov: also, the whole file could be simply mmaped and then chunked any way you like. no need to read it.

s3-helpers/src/upload.rs
```diff
@@ -0,0 +1,286 @@
...
+                }
+            }
+
+            Ok::<(), color_eyre::Report>(())
+        });
+    }
+    drop(progress_tx);
+
+    let mut bytes_so_far: u64 = 0;
+    let mut parts: BTreeMap<NonZeroU16, String> = BTreeMap::new();
```
alekseifedotov, line 183: Why store etags in the tree? Why not an array? the size of the file is known before this code runs.

s3-helpers/src/upload.rs
```diff
@@ -0,0 +1,286 @@
...
+                    let mut lock = ranges.lock().await;
+                    lock.next()
+                };
+                let Some(range) = range_opt else { break };
+
+                let buf = tokio::task::spawn_blocking({
+                    let file = Arc::clone(&file);
+                    move || {
+                        let mut buf = vec![0u8; range.len as usize];
+                        file.read_exact_at(&mut buf, range.start)?;
```
alekseifedotov, line 153: why blocking read?

alekseifedotov: just map the file, save on reads.

s3-helpers/src/upload.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    let mut task_err: Option<color_eyre::Report> = None;
+    while let Some(res) = tasks.join_next().await {
+        if let Err(e) | Ok(Err(e)) = res.map_err(Into::into) {
+            task_err = Some(e);
+            break;
+        }
+    }
+
+    if let Some(e) = task_err {
```
alekseifedotov, line 201: Imagine if one of the upload tasks fails but other tasks keep running. We won't reach this part until every other upload tasks is done. So we will have nothing to abort.

s3-helpers/src/upload.rs
```diff
@@ -0,0 +1,286 @@
...
+                let range_opt = {
+                    let mut lock = ranges.lock().await;
+                    lock.next()
+                };
+                let Some(range) = range_opt else { break };
+
+                let buf = tokio::task::spawn_blocking({
+                    let file = Arc::clone(&file);
+                    move || {
+                        let mut buf = vec![0u8; range.len as usize];
```
alekseifedotov, line 152: allocating a buffer filled with zeros just to immediately overwrite it with data is a crime against performance.

s3-helpers/src/upload.rs
alekseifedotov: In general, it is not a good idea to check for file's metadata and then open it. The file might change or be moved between this two steps. It is better to open it once and then operate on the filedescriptor.

TheButlah: its blocking because tokio's file does not have read_at, but std does. Tokio just uses std under the hood so its not a performance issue. 

Memmapping would indeed work - but then the "blocking" would happen by the OS thread and not by the program whenever reading. we would lose the ability to explicitly see where the blocking happens. Also, we would lose consistency with multipart_download, which is what this PR is based on. What do you think?

s3-helpers/src/upload.rs
TheButlah: marking as resolved to consolidate memmap discussion

s3-helpers/src/upload.rs
TheButlah: Probably. I actually do need the ability to upload a whole directory. Lets do it in the subsequent PR, ill add that functionality there

s3-helpers/src/upload.rs
TheButlah: so true bestie

TheButlah: I do actually need to set zeros, because I am taking &mut [u8]. rust doesn't let me get a mutable ref otherwise.

I could optimize this further with unsafe but I think thats not worth the unsafe.

s3-helpers/src/upload.rs
TheButlah: I'm not sure I understand. The etag is needed for the api call to complete the upload (from what I could see). And why would an array be use.

If you mean why not an array instead of a tree, where the index is the part number - sure yeah. Thats fair.

alekseifedotov: Yeah, it should be an array. Below it creates an array from the btree, that is totally unnecessary.

TheButlah: on second thought, the code would be a lot sloppier if it was an array. I would need to store empty strings or Option<String> in the array. IMO the sparsity of the btree is useful here.

Does it actually matter to you? I can change it but I think its a bit unimportant. Also I don't see why this would impact performance, if thats the concern - this btree is gonna be of length ~200 for a 5GB file. Its not long enough or in a hot enough path for memory pressure or asymptotic complexity to be relevant to the conversation.

s3-helpers/src/upload.rs
TheButlah: I'm not sure thats accurate. There is a `break` statement in the loop, which will terminate us early if any task fails.

s3-helpers/src/upload.rs
TheButlah: fixed :+1:

PR #715, opened 2025-09-12
feat: add devcontainer

alekseifedotov:
Add devcontainer for building orb-software


alekseifedotov:
>lgtm, are you switching away from emacs @alekseifedotov ?

What gives you such impression? Emacs has support for devcontainers.

TheButlah:
Please don't merge something that fundamentally impacts our development flow like this without discussion first?

PR #717, opened 2025-09-15
feat(slot-ctrl): add scratch reg retry counter to diamond

vmenge:
## feat
- reads nvidia's retry counter from sysfs on Diamond

### DIFF
--- a/slot-ctrl/src/test_utils.rs
+++ b/slot-ctrl/src/test_utils.rs
@@ -1,8 +1,9 @@
 use std::fs;
 
 use crate::{
+    domain::ScratchRegRetryCount,
     program::{self, Cli},
-    EfiVarDb, OrbSlotCtrl, RetryCount, RootFsStatus, Slot,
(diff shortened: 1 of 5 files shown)


slot-ctrl/src/test_utils.rs
```diff
@@ -72,6 +82,24 @@ impl Fixture {
...
+
+            fs::write(
+                tempdir
+                    .path()
+                    .join(ScratchRegRetryCount::DIAMOND_COUNT_B_PATH),
+                format!("0x{scratch_reg_retry_count_b}\n"),
+            )
+            .unwrap();
+        }
+
```
pophilpo, line 102: Do we need root access to write to these registers?

vmenge: we just read here in this case, no writing

vmenge: but no root access needed to read

pophilpo: Nvm, I see that we don't write to these values on Diamond, only in test_utls

PR #719, opened 2025-09-16
fix(xtask/deploy): use cargo metadata for systemd svc names

vmenge:
## fix
- use cargo.toml metadata for systemd service names


Qbicz (approved):
A MA ZING

PR #723, opened 2025-09-23
feat(connd): own all connectivity logic

vmenge:
## new
- ensures default `hotspot` wlan profile is created on startup
- ensures default `cellular` wwan profile is created on startup
- imports existing `wpa-supplicant` conf to `NetworkManager` on startup then deletes old file
- smart switching enabled, using url `https://fleet.orb.worldcoin.org/api/v2/online_check` for conn check
- offers dbus methods to
  - apply wifi qr code
  - apply netconfig qr code (complete with orb-app signature verification, cc @max-pankin)
  - check if we have internet connectivity

## changed
- mmcli refactored to be behind `ModemManager` for testing
- `dogstatd::Client` refactored to be behind `StatdClient` for testing
- `iccid`  is optional when reporting to `backend-status`

## related PRs
https://github.com/worldcoin/priv-orb-core/pull/1744
https://github.com/worldcoin/orb-os/pull/1234

##  manual tests
(performed on a dev orb)
- [x] default wlan and wwan profile creation
- [x] wlan conf importing
- [x] smart switching
- [x] applying wifi qr code
- [x] applying netconfig qr code


orb-connd/src/telemetry/net_stats.rs
```diff
@@ -9,30 +10,20 @@ pub struct NetStats {
...
+        let tx_path = stats_path.join("tx_bytes");
+        let rx_path = stats_path.join("rx_bytes");
 
-        let tx_bytes: u64 = lines
-            .next()
-            .ok_or_else(|| eyre!("Missing tx_bytes info line."))?
-            .trim()
-            .parse()?;
+        let tx_bytes = String::from_utf8_lossy(&fs::read(tx_path).await?).parse()?;
+        let rx_bytes = String::from_utf8_lossy(&fs::read(rx_path).await?).parse()?;
```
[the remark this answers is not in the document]
vmenge: image

orb-connd/src/telemetry/net_stats.rs
```diff
@@ -9,30 +10,20 @@ pub struct NetStats {
...
-        let mut lines = output.lines();
+        let tx_path = stats_path.join("tx_bytes");
+        let rx_path = stats_path.join("rx_bytes");
 
-        let tx_bytes: u64 = lines
-            .next()
-            .ok_or_else(|| eyre!("Missing tx_bytes info line."))?
-            .trim()
-            .parse()?;
+        let tx_bytes = String::from_utf8_lossy(&fs::read(tx_path).await?).parse()?;
```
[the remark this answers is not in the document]
vmenge: image

orb-connd/src/service/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+                .add()
+                .await?;
+        }
+
+        Ok(())
+    }
+
+    pub async fn import_wpa_conf(&self, wpa_conf_dir: impl AsRef<Path>) -> Result<()> {
+        let wpa_conf = wpa_conf_dir.as_ref().join("wpa_supplicant-wlan0.conf");
+        match File::open(&wpa_conf).await {
```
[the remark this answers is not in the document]
vmenge: stupid clanker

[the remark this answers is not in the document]
vmenge: image

orb-connd/src/service/wifi.rs
```diff
@@ -0,0 +1,48 @@
...
+            .wrap_err_with(|| format!("missing wifi ssid on qr code: {wifi_str}"))?
+            .to_string();
+
+        let hidden = map
+            .get("H")
+            .map(|h| h.parse())
+            .transpose()?
+            .unwrap_or_default();
+
+        let psk = map.get("P").map(|p| p.to_string());
```
valff: Why not keep using the existing `nom`-based Mecard parser for this?

orb-connd/src/service/netconfig.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct NetConfig {
+    pub wifi_credentials: Option<wifi::Credentials>,
+    pub wifi_enabled: Option<bool>,
+    pub smart_switching: Option<bool>,
+    pub airplane_mode: Option<bool>,
+    pub created_at: DateTime<Utc>,
+}
+
+impl NetConfig {
+    pub fn parse(netconfig_str: &str) -> Result<NetConfig> {
```
valff: Have you considered using `nom` crate for this kind of parser?

orb-connd/src/service/netconfig.rs
```diff
@@ -0,0 +1,205 @@
...
+        let netconfig_str = netconfig_str.replace("WIFI:", "");
```
valff: What if there is an SSID named `MYWIFI:HOME`? A real parser wouldn't have this issue.

orb-connd/src/service/netconfig.rs
```diff
@@ -0,0 +1,205 @@
...
+        let map: HashMap<_, _> = netconfig_str
+            .split(";")
+            .flat_map(|entry| entry.split_once(":"))
+            .collect();
```
valff: If there is an entry without a colon, we ignore it. Perhaps this is fine.
But what if there is a password that contains a colon? This is quite common.

valff: Ok, a password with a colon would pass, because we use `split_once` here, and it would be something like this: `P:abc:def`

valff: But what if the password contains `;`? What if it contains `\"`? What if it's quoted?

orb-connd/src/service/wifi.rs
vmenge: switched back to nom parser, but removed hex encoding for ssids (kept it for pws, but changed check to make sure psks are 64 chars to avoid false positives)

orb-connd/src/service/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+
+            Ok(())
+        }
+        .await
+        .inspect_err(|e| error!("failed to apply netconfig qr with {e}"))
+    }
+
+    async fn apply_magic_reset_qr(&self) -> ZResult<()> {
+        info!("trying to apply magic reset qr");
+        self.nm.set_wifi(false).await.into_z()?;
```
AlexKaravaev, line 439: This looks scary a bit, do we really need to **disable** wifi ?

AlexKaravaev: maybe wrap in clojure to reenable wifi always in case error happens ?

orb-connd/src/service/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+            // Pearl orbs do not support extra NetConfig fields
+            if self.platform == OrbOsPlatform::Pearl {
+                return Ok(());
+            }
+
+            if let Some(_airplane_mode) = netconf.airplane_mode {
+                warn!("airplane mode is not supported yet!");
+            }
+
+            if let Some(wifi_enabled) = netconf.wifi_enabled {
```
AlexKaravaev, line 413: If the orb is wifi-only, do we want to disable wifi via netconfig ? maybe only apply if it's lte ?

orb-connd/src/lib.rs
```diff
@@ -1,71 +1,91 @@
...
-        tasks.push(
-            cellular::start(Duration::from_secs(30), Duration::from_secs(20)).await,
-        )
+    if let Err(e) = connd.import_wpa_conf(wpa_conf_dir).await {
+        warn!("failed to import legacy wpa config {e}");
     }
 
-    tasks.extend(telemetry::start().await?);
+    if let Err(e) = connd.ensure_networking_enabled().await {
+        warn!("failed to ensure networking is enabled {e}");
```
AlexKaravaev, line 52: hm, should we crash here and systemd would restart the service ?

orb-connd/src/service/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+            info!("dbus service spawned successfully!");
+            futures::future::pending::<()>().await;
+
+            Ok(())
+        })
+    }
+
+    pub async fn ensure_networking_enabled(&self) -> Result<()> {
+        if !self.nm.networking_enabled().await? {
+            self.nm.set_networking(true).await?;
```
AlexKaravaev, line 81: should we attempt to enable networking every 5 minutes or so in a background ?

orb-connd/src/service/mod.rs
vmenge: i think i can remove the disabling of wifi, will do so in a next pr

orb-connd/src/service/mod.rs
vmenge: oh, thats a good point. i think at the time i thought we wouldn't have Diamond EVTs in the field but we do so :sweat_smile: 

will change in a followup pr as well :+1:

orb-connd/src/lib.rs
vmenge: no i think its fine tbh, the only reason this would error out really is if we can't talk to NM over dbus, or if NM is in a weird state, by which point i think we'll have failed before this 

this error should never really occur, but if it does its weird enough that i added the logging and am assuming the orb will have connectivity so i dont want the error to block connd from starting (it might do so when connd can  start without issues)

orb-connd/src/service/mod.rs
vmenge: i think its not needed -- right now there is nothing that turns networking off, and if for some reason it is turned off a power cycle would fix it

PR #728, opened 2025-10-02
feat: Add support for new jobs commands - gimbal recalibration and beacon

matthewpaul:
This pull request introduces several new job handlers and significant improvements to the `orb-jobs-agent` crate, focusing on device management and calibration functionality. Notably, it adds new handlers for beaconing, MCU information, reading and resetting gimbal calibration, and refactors the reboot logic to be more modular and reusable. These changes enhance the agent's ability to manage and update device state, especially for calibration workflows on Pearl devices.

**New Job Handlers and Features:**

* Added new `reset_gimbal` job handler, which backs up, updates, and validates the gimbal calibration file, schedules a reboot via D-Bus, and only supports Pearl devices. This handler ensures calibration values are set to the desired offsets and provides detailed progress updates.
* Implemented `read_gimbal` handler to read and validate the gimbal calibration JSON file, returning its contents as job output.
* Introduced `beacon` handler to execute the `orb-beacon` command and return its output, handling errors gracefully.
* Added `mcu` handler to execute `orb-mcu-util info` with validation for unsupported arguments, returning structured job results.

**Reboot Flow Refactor:**

* Refactored reboot logic into a reusable `run_reboot_flow` function and `RebootPlan` struct, enabling other handlers (like `reset_gimbal`) to leverage the reboot flow with custom progress and error handling. [[1]](diffhunk://#diff-d846301a3ebfefbe05541184a98b1890a19e1452c54c0ed5638371df52967c25L7-R81) [[2]](diffhunk://#diff-d846301a3ebfefbe05541184a98b1890a19e1452c54c0ed5638371df52967c25R97-R98) [[3]](diffhunk://#diff-d846301a3ebfefbe05541184a98b1890a19e1452c54c0ed5638371df52967c25R108-L78)

**Integration and Registration:**

* Registered the new handlers (`beacon`, `reset_gimbal`) in the job handler builder and module system, making them available for use. [[1]](diffhunk://#diff-828c40d1bb4e8a2bc1a35351f4a0549510dc8c0b004035a96277f1a83096a409R1-R9) [[2]](diffhunk://#diff-6ef7e04362a7d78e147b3e7ed0702028b79c0fe368c0284d1ba85f96b0664f28L3-R4) [[3]](diffhunk://#diff-6ef7e04362a7d78e147b3e7ed0702028b79c0fe368c0284d1ba85f96b0664f28R37-R41)

**Dependency Updates:**

* Added `chrono` as a workspace dependency to support time-based operations in new handlers.


orb-jobs-agent/src/handlers/reset_gimbal.rs
```diff
@@ -0,0 +1,214 @@
...
+        fs::metadata(&calibration_path).await.with_context(|| {
+            format!("calibration file not found: {}", calibration_path.display())
+        })?;
+
+        let backup_path = create_backup(&calibration_path).await?;
+        info!(?backup_path, "created calibration backup");
+
+        let updated_calibration = update_calibration_file(&calibration_path).await?;
+
+        let scheduled_micros = compute_reboot_deadline()?;
```
AlexKaravaev: do we need to reboot after reset ? I am not that familiar for some reason  Ithought we only need orb-core restart

orb-jobs-agent/src/handlers/beacon.rs
```diff
@@ -0,0 +1,37 @@
...
+use color_eyre::{eyre::WrapErr, Result};
+use orb_relay_messages::jobs::v1::JobExecutionUpdate;
+
+/// command format: `beacon`
+#[tracing::instrument(skip(ctx))]
+pub async fn handler(ctx: Ctx) -> Result<JobExecutionUpdate> {
+    let output = ctx
+        .deps()
+        .shell
+        .exec(&["orb-beacon"])
```
AlexKaravaev, line 14: btw in the future we can also add flashing animation of the ring(currently it's only sound)

orb-jobs-agent/src/handlers/reset_gimbal.rs
matthewpaul: I implemented the reboot since this is exactly what the tech support team currently does for the flow. I expect that it's sufficient to simply reboot orb-core, although I'm not sure if this service needs any specific permissions to reboot services. 

If we want to go down that route, I would be interested in an additional job for restarting a preset list of services. I think it could be really useful to be able to remotely restart `worldcoin-core`, `worldcoin-jobs-agent`, `worldcoin-update-agent`, maybe some others? WDYT?

orb-jobs-agent/src/handlers/beacon.rs
matthewpaul: That would be great :) this feature will be really useful I think

orb-jobs-agent/src/handlers/reset_gimbal.rs
AlexKaravaev: yeah sgtm ! I think we definitely need to add restart for update-agent for better UX - we discussed it sometime ago that otherwise when user presses update button there might be a delay

AlexKaravaev (approved):

orb-jobs-agent/src/handlers/beacon.rs
```diff
@@ -0,0 +1,37 @@
...
+        .await
+        .wrap_err("failed to spawn orb-beacon")?
+        .wait_with_output()
+        .await
+        .wrap_err("failed to get orb-beacon output")?;
+
+    if !output.status.success() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        return Ok(ctx.failure().stderr(stderr));
```
vmenge: you can also just return an error, will have the same effect and be a bit more idiomatic 

after factor, all failures are gracefully handled and sent back as an error to the backend

orb-jobs-agent/src/handlers/mcu.rs
```diff
@@ -1,13 +1,33 @@
...
+        .exec(&["orb-mcu-util", "info"])
+        .await
+        .context("failed to spawn orb-mcu-util info")?
+        .wait_with_output()
+        .await
+        .context("failed to wait for orb-mcu-util info")?;
+
+    if !output.status.success() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        return Ok(ctx.failure().stderr(stderr));
```
vmenge: same thing as previous comment
you can actuall use `bail!` macro as well :D

orb-jobs-agent/src/handlers/reboot.rs
```diff
@@ -4,14 +4,81 @@ use color_eyre::{
...
+            ..Self::default()
+        }
+    }
+}
+
+pub async fn run_reboot_flow<F, Fut>(
+    ctx: Ctx,
+    command_label: &str,
+    on_reboot: F,
+) -> Result<JobExecutionUpdate>
```
vmenge, line 47: tbh while i enjoy callbacks i think using one makes a  bit harder to reason about things here, even though i realize you have need for the function elsewhere

not requesting for changes in this part of the code in particular though splitting some of the reboot logic into re-usable code could help solve the issue

orb-jobs-agent/src/handlers/reboot.rs
```diff
@@ -36,46 +105,32 @@ pub async fn handler(ctx: Ctx) -> Result<JobExecutionUpdate> {
...
 
-            RebootStatus::write_pending_lockfile(store_path, ctx.execution_id())
-                .await?;
-
-            ctx.progress()
-                .stdout("rebooting\n")
-                .send()
-                .await
-                .map_err(|e| eyre!("failed to send progress {e:?}"))?;
+            let plan = on_reboot(ctx.clone()).await?;
```
vmenge, line 82: this won't work unfortunately :(

closure passed to `on_reboot` will shutdown the orb with `shutdown now`, so none of the code after this will run 

unfortunately tests won't catch it as only shell commands run inside the container, so the actual impl will keep running unlike in a real orb :cry: 

would make more sense to move the `orb-mcu-util reboot orb`  and `shutdown now` to outside of the closure

orb-jobs-agent/src/handlers/reset_gimbal.rs
vmenge: job for restarting a pre-selected list of services will come soon TM

either when im back from vacation or if someone else does it first :P

orb-jobs-agent/src/handlers/reset_gimbal.rs
```diff
@@ -0,0 +1,214 @@
...
+        schedule_reboot(scheduled_micros).await?;
```
vmenge: reboot command should be moved to the `run_reboot_flow` fn, and that should be the `orb-mcu-util reboot orb` followed by `sudo shutdown now` from the reboot handler :pray: 

supervisor reboot shouldnt be used :sweat_smile: 
it fucks up the nvidia retry counter

orb-jobs-agent/src/handlers/reset_gimbal.rs
```diff
@@ -0,0 +1,214 @@
...
+fn compute_reboot_deadline() -> Result<u64> {
+    let deadline = Utc::now() + Duration::seconds(1);
+    deadline
+        .timestamp_micros()
+        .try_into()
+        .context("scheduled shutdown time is before UNIX epoch")
+}
+
+async fn schedule_reboot(micros: u64) -> Result<()> {
+    let connection = Connection::session()
```
TheButlah: dbus connections should be passed in via dependency injection all the way from main, not constructed locally. Otherwise its impossible to test.

As an example of what this tends to look like, see [here](https://github.com/worldcoin/orb-software/blob/21380afb1baa6f622f07f0a14d7bdc40fa62d5da/supervisor/tests/it/helpers.rs#L17-L88). You only need to inject the connection.

orb-jobs-agent/src/handlers/reset_gimbal.rs
```diff
@@ -0,0 +1,214 @@
...
+
+#[proxy(
+    interface = "org.worldcoin.OrbSupervisor1.Manager",
+    default_service = "org.worldcoin.OrbSupervisor1",
+    default_path = "/org/worldcoin/OrbSupervisor1/Manager"
+)]
+trait SupervisorManager {
+    #[zbus(name = "ScheduleShutdown")]
+    fn schedule_shutdown(&self, kind: &str, when: u64) -> zbus::Result<()>;
+}
```
TheButlah: we should *ideally* follow the pattern where we have a orb-supervisor-dbus crate that implements a proxy and empty interface that can have its implementation injected. See [here](https://github.com/worldcoin/orb-software/blob/main/attest/dbus/src/lib.rs) for example of how to define a -dbus crate, [here](https://github.com/worldcoin/orb-software/blob/21380afb1baa6f622f07f0a14d7bdc40fa62d5da/orb-info/src/orb_token.rs#L86-L151) as an example of how to implement and use it.

I would understand if you think its a bit out of scope though.

AlexKaravaev (approved):
LGTM! One comment is that reboot on pearl is still not supported, we are trying ti push the fw release as hard as we can.

And also I would still like more Rust-minded people @vmenge @TheButlah @sfikastheo to take a look if needed

nix/machines/worldcoin-hil-munich-5/configuration.nix
```diff
@@ -0,0 +1,14 @@
+# Edit this configuration file to define what should be installed on
+# your system.  Help is available in the configuration.nix(5) man page
+# and in the NixOS manual (accessible by running ‘nixos-help’).
+
+{ config, pkgs, lib, ... }:
```
TheButlah, line 5: lets try to avoid combining unrelated changes in the same pr

TheButlah (changes requested):
looks good, just lets remove the nix configuration stuff. FWIW the code, especially the traits, does feel very OOP enterprise-y which generally we try to avoid, but maybe it was just copying similar patterns that exist in jobs agent? Im missing some context there so im not going to push one way or the other.

orb-jobs-agent/src/handlers/reset_gimbal.rs
```diff
@@ -0,0 +1,166 @@
...
+            format!("calibration file not found: {}", calibration_path.display())
+        })?;
+
+        let backup_path = create_backup(&calibration_path).await?;
+        info!(?backup_path, "created calibration backup");
+
+        let updated_calibration = update_calibration_file(&calibration_path).await?;
+
+        let response = serde_json::json!({
+            "backup": backup_path
```
TheButlah: shouldn't this be strongly typed instead?

orb-jobs-agent/src/handlers/reset_gimbal.rs
```diff
@@ -0,0 +1,166 @@
...
+    let contents = fs::read_to_string(calibration_path)
+        .await
+        .with_context(|| {
+            format!(
+                "failed to read calibration file at {}",
+                calibration_path.display()
+            )
+        })?;
+
+    let mut calibration: Value = serde_json::from_str(&contents)
```
TheButlah: use a proper struct not the dynamic type

nix/machines/worldcoin-hil-munich-5/configuration.nix
matthewpaul: I'm not sure why this is showing up here, I haven't touched this file as far as I can tell

PR #734, opened 2025-10-07
Pophilpo/hil ota

pophilpo:


hil/src/ssh_wrapper.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        session
+            .userauth_password(&username, &password)
+            .map_err(|e| {
+                color_eyre::eyre::eyre!("SSH password authentication failed: {}", e)
+            })?;
+
+        if !session.authenticated() {
+            bail!("SSH authentication failed");
+        }
```
TheButlah: you're doing a ton of blocking work in async. Never ever block in async, use tokio::task::spawn_blocking

hil/src/ssh_wrapper.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        channel
+            .exec(command)
+            .map_err(|e| color_eyre::eyre::eyre!("Failed to execute command: {}", e))?;
+
+        let mut stdout = String::new();
+        let mut stderr = String::new();
+
+        channel
+            .read_to_string(&mut stdout)
```
TheButlah: blocks in async.

hil/src/ssh_wrapper.rs
```diff
@@ -0,0 +1,127 @@
+use color_eyre::{eyre::bail, Result};
+use ssh2::Session as Ssh2Session;
```
TheButlah: instead of ssh2, which requires you to dynamically link against C libraries, why not just use https://docs.rs/cmd_lib/latest/cmd_lib/ (blocking but terse) or https://docs.rs/tokio/latest/tokio/process/struct.Command.html (async but verbose) and the actual `ssh` CLI.

Alternatively why not https://github.com/Eugeny/russh which actually does async in native rust

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,634 @@
...
+    /// Maximum reconnection attempts after reboot
+    #[arg(long, default_value = "50")]
+    max_reconnect_attempts: u32,
+
+    /// Sleep time between reconnection attempts in seconds
+    #[arg(long, default_value = "5")]
+    reconnect_sleep_secs: u64,
+
+    /// Serial port path for boot log capture
+    #[arg(long, default_value = "/dev/ttyUSB1")]
```
TheButlah: do not default this, its too hardware-dependent.

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,634 @@
...
+    serial_path: PathBuf,
```
TheButlah: Add another argument called `serial-id` and put it in a [group](https://github.com/clap-rs/clap/blob/c9a39a534c3e95926be272765bec48a80e5ea9e7/examples/tutorial_derive/04_03_relations.rs#L10) called serial. This will make it so that the two flags are mutually exclusive, you can either provide serial by path or by its id.

You can then open /dev/serial/by-id/yourid

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,634 @@
...
+
+        if let Some(captures) = slot_regex.captures(&result.stdout) {
+            let slot_letter = captures.get(1).unwrap().as_str();
+            let slot_name = format!("slot_{slot_letter}");
+            info!("Current slot: {}", slot_name);
+            Ok(slot_name)
+        } else {
+            bail!("Could not parse current slot from: {}", result.stdout);
+        }
+    }
```
TheButlah, line 293: nit: This and a lot of the methods could be a free function rather than an associated function (aka method). Because it doesn't use much at all from `self`.

I find that it is more verbose but more explicit and less OOP to try to use free-standing functions when possible.

Good rule of thumb - if you start to need more than two of the fields of `self`, then `self` is ok to use (or consider breaking the fields up into logical containers, for example all of your ssh dialing info (user, password, port, address) could be a SshDialInfo struct.

hil/src/ssh_wrapper.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::io::Read;
+use std::net::TcpStream;
+use tracing::{debug, info};
+
+/// SSH wrapper that supports both password and key authentication
+/// Will be removed in favor of ssh keys
+pub struct SshWrapper {
```
TheButlah, line 31: Generally you should always `#[derive(Debug)]`

hil/src/ssh_wrapper.rs
```diff
@@ -0,0 +1,127 @@
...
+    session: Ssh2Session,
+}
+
+impl SshWrapper {
+    pub async fn connect(
+        host: String,
+        port: u16,
+        username: String,
+        password: String,
```
TheButlah: given that you use this same set of arguments in several functions, maybe it makes sense to create a simple SshConnectArgs struct.

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,634 @@
...
+
+        if !result.is_success() {
+            bail!("wipe_overlays function failed: {}", result.stderr);
+        }
+
+        info!("Overlays wiped successfully");
+        Ok(())
+    }
+
+    #[instrument(skip(self, session))]
```
TheButlah: when you are skipping all arguments, you can just say `skip_all`

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,634 @@
...
+}
+
+#[derive(Debug, Clone, clap::ValueEnum)]
+enum Platform {
+    Diamond,
+    Pearl,
+}
+
+impl Ota {
+    #[instrument(skip(self))]
```
TheButlah:
```suggestion
```
I wouldn't skip instrumenting the arguments here. Its good to know what the command was run with when seeing the traces.

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,634 @@
...
+            }
+
+            let _ = kill_tx.send(());
+            Ok::<(), color_eyre::Report>(())
+        };
+
+        let ((), ()) = tokio::try_join! {
+            boot_log_fut,
+            async {
+                let _ = reader_task.await.wrap_err("serial reader task panicked")?;
```
TheButlah: what is this discarding and why?

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,634 @@
...
+        info!("Running orb-update-verifier");
+
+        let result = session
+            .execute_command("TERM=dumb sudo orb-update-verifier")
+            .await
+            .wrap_err("Failed to run orb-update-verifier")?;
+
+        if !result.is_success() {
+            bail!("orb-update-verifier failed: {}", result.stderr);
+        }
```
TheButlah: shorcut for `if condition{ bail!() }` is `ensure!()`

TheButlah (changes requested):
main problems:
* I dont think we should be using ssh2 crate due to the c dependency
* You block in async many many times in many places. Please go through the code and fix those then I will re-review

hil/src/ssh_wrapper.rs
pophilpo: Switched to using the Command variant, this requires the caller to have sshpass installed for password auth though

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,714 @@
...
+use tokio_stream::wrappers::BroadcastStream;
+use tracing::{debug, error, info, instrument, warn};
+
+/// Over-The-Air update command for the Orb
+#[derive(Debug, Parser)]
+#[command(group = clap::ArgGroup::new("serial").required(true).multiple(false))]
+pub struct Ota {
+    /// Target version to update to
+    #[arg(long)]
+    version: String,
```
TheButlah: Rename to version_to or target_version

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,714 @@
...
+
+    /// Hostname of the Orb device
+    #[arg(long)]
+    host: String,
```
TheButlah: Rebame to hostname

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,714 @@
...
+    /// Username
+    #[arg(long, default_value = "worldcoin")]
+    username: String,
+
+    /// Password for authentication (mutually exclusive with --key-path)
+    #[arg(long, required_unless_present = "key_path")]
+    password: Option<String>,
+
+    /// Path to SSH private key for authentication (mutually exclusive with --password)
+    #[arg(long, required_unless_present = "password")]
```
TheButlah: Cant you use group here too? Instead of the required thing

TheButlah (commented):
E

hil/src/commands/ota.rs
```diff
@@ -0,0 +1,714 @@
...
+    async fn connect(&self) -> Result<SshWrapper> {
+        info!("Connecting to Orb device at {}:{}", self.host, self.port);
+
+        let auth = match (&self.password, &self.key_path) {
+            (Some(password), None) => AuthMethod::Password(password.clone()),
+            (None, Some(key_path)) => AuthMethod::Key {
+                private_key_path: key_path.clone(),
+            },
+            (Some(_), Some(_)) => {
+                bail!("Cannot specify both --password and --key-path");
```
TheButlah: Change to unreachable!() And combine with subsequent match branch Since we already enforce via clap

TheButlah (approved):
Very good job. Thanks @pophilpo !

hil/src/commands/ota.rs
pophilpo: Good point, implemented that

PR #747, opened 2025-10-23
feat(services): Touching service files

sfikastheo:
Removing `backend-connect` & touching up the rest of the services for NITs

### DIFF
--- a/attest/debian/orb-attest.worldcoin-attest.service
+++ b/attest/debian/orb-attest.worldcoin-attest.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Short lived attestation token daemon
-## persistent contains SE pairing keys
+After=worldcoin-configure-can@can0.service
+Wants=worldcoin-configure-can@can0.service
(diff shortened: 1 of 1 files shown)


attest/debian/orb-attest.worldcoin-attest.service
```diff
@@ -2,8 +2,8 @@
 Description=Short lived attestation token daemon
 ## persistent contains SE pairing keys
 After=usr-persistent.mount
-Requires=worldcoin-backend-online.target
```
AlexKaravaev, line 5: Not 100%, but worldcoin-backend-online probably required network-online which required can interface to be up. Dbus as far as I have looked doesn't - means we can in theory go into situation where can is not up while attest is up. I think it shouldn't break anything since attest would retry, but just wanted to mention

sfikastheo: true edge-case, not sure about the details at all, but added a dependency for can

PR #903, opened 2026-01-05
fix(ota-hil): set recovery mode before capture logs

chrisgalanis:
This pull request introduces several improvements and enhancements to the OTA update and reboot process, focusing on reliability, observability, and configurability. Key changes include more robust time synchronization checks, improved handling of boot log capture, and better control over the reboot sequence to ensure devices do not inadvertently enter recovery mode.

**OTA Update Flow Improvements:**

* Added a new `--skip-time-sync-before-reboot` flag to allow skipping the NTP time synchronization check before the first reboot, with clear logging and fallback behavior. [[1]](diffhunk://#diff-4417a067e971532511cd378570394d24d393625d1e510ecc9475239495917805R72-R76) [[2]](diffhunk://#diff-4417a067e971532511cd378570394d24d393625d1e510ecc9475239495917805L108-R133)
* Created log directories as needed before OTA operations, improving robustness when writing logs.

**Reboot and Recovery Handling:**

* The reboot process now waits for SSH disconnection before manipulating the recovery pin, ensuring the pin is held only during the actual shutdown window. The duration for holding the recovery pin is increased to 20 seconds, and a short delay is added after releasing the pin to allow for USB device re-enumeration.
* The logic for setting the recovery pin during reboot is now parameterized based on the recovery state, improving clarity and correctness. [[1]](diffhunk://#diff-201f44f16c361cb2807a9ba916d437dd592b40a19e0a1b9f095b37e3ea38cd7cL40-R45) [[2]](diffhunk://#diff-201f44f16c361cb2807a9ba916d437dd592b40a19e0a1b9f095b37e3ea38cd7cL49-R54)
* Increased the sleep duration after reboot from 4 to 10 seconds to provide more time for device stabilization.

**Boot Log Capture Enhancements:**

* Boot log capture is now performed in a background task, running concurrently with SSH reconnection attempts. The capture waits until a login prompt is detected, writes data incrementally to disk, and provides detailed logging on success or failure. [[1]](diffhunk://#diff-ea3806106fed5cd38c7fb89471973b5dd24656ee7c4c3dc99b3ba9043eb7cfe7R155-R188) [[2]](diffhunk://#diff-ea3806106fed5cd38c7fb89471973b5dd24656ee7c4c3dc99b3ba9043eb7cfe7L154-R261) [[3]](diffhunk://#diff-ea3806106fed5cd38c7fb89471973b5dd24656ee7c4c3dc99b3ba9043eb7cfe7R281-R308)
* The boot log capture function is refactored into a static method for easier background invocation and improved testability.

**Time Synchronization Robustness:**

* The time synchronization check now automatically detects and uses either `timedatectl` or `chronyc`, with command timeouts and more robust output parsing. If neither tool is found, a clear error is raised.
* Time sync is now always checked after reboot and before starting the update, regardless of the pre-reboot check configuration.

**Utility and Observability Improvements:**

* Added a utility method to wait for SSH disconnection, improving reliability in detecting device shutdown before proceeding with hardware-level operations.

These changes collectively improve the reliability, observability, and flexibility of the OTA update and device reboot process, making it easier to diagnose issues and adapt to different environments.

---

**OTA update and logging enhancements:**
- Added `--skip-time-sync-before-reboot` flag to optionally skip NTP check before the first reboot, with clear logging and fallback. [[1]](diffhunk://#diff-4417a067e971532511cd378570394d24d393625d1e510ecc9475239495917805R72-R76) [[2]](diffhunk://#diff-4417a067e971532511cd378570394d24d393625d1e510ecc9475239495917805L108-R133)
- Ensured log directories are created before writing logs, improving robustness.

**Reboot sequence and recovery pin handling:**
- Waits for SSH disconnect before holding the recovery pin, increases hold duration to 20 seconds, and adds a delay after releasing the pin for USB re-enumeration.
- Parameterized recovery pin state during reboot for clarity and correctness. [[1]](diffhunk://#diff-201f44f16c361cb2807a9ba916d437dd592b40a19e0a1b9f095b37e3ea38cd7cL40-R45) [[2]](diffhunk://#diff-201f44f16c361cb2807a9ba916d437dd592b40a19e0a1b9f095b37e3ea38cd7cL49-R54)
- Increased post-reboot sleep from 4 to 10 seconds for device stabilization.

**Boot log capture improvements:**
- Boot log capture runs in the background, waits for login prompt, writes incrementally to disk, and provides detailed success/failure logs. [[1]](diffhunk://#diff-ea3806106fed5cd38c7fb89471973b5dd24656ee7c4c3dc99b3ba9043eb7cfe7R155-R188) [[2]](diffhunk://#diff-ea3806106fed5cd38c7fb89471973b5dd24656ee7c4c3dc99b3ba9043eb7cfe7L154-R261) [[3]](diffhunk://#diff-ea3806106fed5cd38c7fb89471973b5dd24656ee7c4c3dc99b3ba9043eb7cfe7R281-R308)
- Refactored boot log capture into a static method for better background usage.

**Time synchronization checks:**
- Automatically detects and uses `timedatectl` or `chronyc` for time sync, with timeouts and robust output parsing. Raises error if neither is found.
- Always checks time sync after reboot, regardless of pre-reboot check.

**Utility and reliability improvements:**
- Added utility to wait for SSH disconnection, improving reliability in detecting shutdown before hardware operations.


hil/src/commands/ota/mod.rs
```diff
@@ -94,6 +99,14 @@ impl Ota {
         let _start_time = Instant::now();
         info!("Starting OTA update to version: {}", self.target_version);
 
+        // Create log directory if it doesn't exist
```
AlexKaravaev: i think we can remove comments like this one, they don't bring much value

hil/src/commands/ota/mod.rs
```diff
@@ -105,7 +118,19 @@ impl Ota {
                 system::wipe_overlays(&session).await.inspect_err(|e| {
                     error!("Failed to wipe overlays: {}", e);
                 })?;
-                info!("Overlays wiped successfully, rebooting device");
+                info!("Overlays wiped successfully");
+
+                if !self.skip_time_sync_before_reboot {
```
AlexKaravaev, line 134: why we need this I wonder ?

hil/src/commands/ota/reboot.rs
```diff
@@ -20,30 +20,51 @@ impl Ota {
...
-
-        // Wait for recovery pin task to complete
         recovery_task
             .await
             .wrap_err("recovery pin task panicked")??;
 
+        // Brief delay to allow USB device to be re-enumerated and udev rules to apply
+        // after FTDI GPIO is released. The FTDI device detaches/reattaches kernel
+        // drivers which causes /dev/ttyUSB* to be recreated.
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
AlexKaravaev: timings like this might be weak spot and break often - but if works now i am fine

hil/src/commands/ota/reboot.rs
```diff
@@ -103,27 +152,40 @@ impl Ota {
         );
     }
 
+    /// Captures boot logs from serial port in the background
     #[instrument(skip_all)]
-    async fn capture_boot_logs(&self, log_suffix: &str) -> Result<()> {
-        let platform_name = format!("{:?}", self.platform).to_lowercase();
+    async fn capture_boot_logs_static(
```
AlexKaravaev: i think it's bad practice to have static in the end of the name

hil/src/commands/ota/reboot.rs
```diff
@@ -185,36 +278,34 @@ impl Ota {
...
                     }
                     Err(_) => {
+                        // Timeout on reading - continue waiting
                         continue;
                     }
                 }
             }
 
-            if start_time.elapsed() >= timeout && !found_login_prompt {
+            if found_login_prompt {
```
AlexKaravaev, line 306: where do we set timeout for this?

hil/src/commands/ota/system.rs
```diff
@@ -138,24 +138,91 @@ pub async fn wait_for_time_sync(session: &SshWrapper) -> Result<()> {
...
+        } else {
+            "chronyc"
+        }
+    );
+
     for attempt in 1..=MAX_ATTEMPTS {
-        let result = session
-            .execute_command("TERM=dumb timedatectl status")
+        // Timeout for individual command execution (10 seconds is generous for timedatectl/chronyc)
+        const COMMAND_TIMEOUT: Duration = Duration::from_secs(10);
```
AlexKaravaev: better to define it in top of the function for readibility

hil/src/boot.rs
```diff
@@ -74,7 +79,7 @@ pub async fn reboot(recovery: bool, device: Option<&FtdiId>) -> Result<()> {
     })
     .await
     .wrap_err("task panicked")??;
-    tokio::time::sleep(Duration::from_secs(4)).await;
+    tokio::time::sleep(Duration::from_secs(10)).await;
```
AlexKaravaev: move to the top of the file or maybe have as separate arg to cli so you don't tune it in sources every time?

AlexKaravaev (changes requested):

hil/src/commands/ota/mod.rs
chrisgalanis: So, basically the problem was with the worldcoin-key-retrival service. With pearl, teh harware clock is not working so if it NTP server is not synced then worldcoin-attest is failing and also key-retrieval. This should be fixed with new key-retrieval fix, that is why i ketp it as a flag, so we can test older commits, that don't have key-retrival fixed

hil/src/commands/ota/reboot.rs
chrisgalanis: I have tested and look reliable enough. Even though if we don't ahve logs in the future we will now the root of failure and fix it.

hil/src/commands/ota/reboot.rs
chrisgalanis: The timeout is the basically if the login prompth is fouund or not. If login prompt is never found, then the timeout of ssh will fail later so the process will exit with error

hil/src/commands/ota/reboot.rs
chrisgalanis: I will fix!

hil/src/commands/ota/system.rs
chrisgalanis: I will fix

hil/src/boot.rs
chrisgalanis: I will fix!

hil/src/commands/ota/reboot.rs
AlexKaravaev: Better then make it a constant

hil/src/commands/ota/mod.rs
sfikastheo: I am also not getting it sorry. 
> With pearl, teh harware clock is not working so if it NTP server is not synced then worldcoin-attest is failing and also key-retrieval.

That is not true. The clock will synchronize and https requests is going to succeed. And the delay is not big for that

chrisgalanis: By default this is not correct. Because reboot is manually triggered faster than the NTP server syncrhonizes the clock....

sfikastheo (approved):

hil/src/commands/ota/reboot.rs
```diff
@@ -15,35 +15,59 @@ use tracing::{debug, error, info, instrument, warn};
 
 use super::Ota;
 
+const DELAY_CAPTURE_LOGS: u64 = 200;
```
AlexKaravaev, line 18: nit: better to specify MS here as it's not clear what 200 is (seconds, ms, etc.)

hil/src/commands/ota/system.rs
```diff
@@ -134,28 +134,93 @@ pub async fn wait_for_time_sync(session: &SshWrapper) -> Result<()> {
 
     const MAX_ATTEMPTS: u32 = 60; // 60 attempts = 2 minutes max wait
     const SLEEP_DURATION: Duration = Duration::from_secs(2);
+    // Timeout for individual command execution (10 seconds is generous for timedatectl/chronyc)
```
AlexKaravaev, line 137: nit: (10 seconds is generous for timedatectl/chronyc) is not helpful as a comment

hil/src/boot.rs
```diff
@@ -74,7 +82,7 @@ pub async fn reboot(recovery: bool, device: Option<&FtdiId>) -> Result<()> {
...
```
AlexKaravaev: You are not using the const here

AlexKaravaev: this is only blocknig comment - everything else is great job!

AlexKaravaev (changes requested):

PR #1017, opened 2026-02-24
feat(ui): Analysis UI

chrisgalanis:
This pull request introduces a new mechanism for selecting operator default colors based on the current release type (such as "Analysis" or "Prod") for both Pearl and Diamond orb types. It achieves this by reading the OS release type at runtime and updating the logic throughout the codebase to use the appropriate color constants. Additionally, new color constants specific to the "Analysis" release type are added.

**Release-type aware color selection:**

* Added a global static `RELEASE_TYPE` (using `LazyLock`) in `ui/src/main.rs` that reads the current `OrbRelease` from `/etc/os-release`, defaulting to `Prod` if unavailable.
* Introduced helper functions `pearl_operator_default()` and `diamond_operator_default()` in `ui/src/engine/mod.rs` to select the correct color constant based on `RELEASE_TYPE`.
* Updated all usages of `PEARL_OPERATOR_DEFAULT` and `DIAMOND_OPERATOR_DEFAULT` in operator logic (idle, pulse, signup_phase, and runner modules) to use the new helper functions, ensuring color selection is dynamic based on release type. [[1]](diffhunk://#diff-72c9cab3257c418720ce42498122a095592a4776345521b342057e002eaa2d03L69-R70) [[2]](diffhunk://#diff-72c9cab3257c418720ce42498122a095592a4776345521b342057e002eaa2d03L144-R147) [[3]](diffhunk://#diff-6d48d3772a75dd86e50a1cbe8bf12a28cffbd8300439ba55151d6ee01867b86dL42-R43) [[4]](diffhunk://#diff-c026af17686ffa346aa7a6a031f79bbcd4b8d8f262a2d77b86c19981727dcf8cL242-R246) [[5]](diffhunk://#diff-610f68ac1137c6887e40f32518fcc215f4b91117c226b7e47d43effb289cefddL81-R81) [[6]](diffhunk://#diff-a5573cbfc9aa9a32f73de88de17640aec71d0a0428bf9c0156eb052903352acdL78-R78)

**New color constants:**

* Added `PEARL_OPERATOR_ANALYSIS` and `DIAMOND_OPERATOR_ANALYSIS` color constants to the `Argb` implementation for use in the "Analysis" release type. [[1]](diffhunk://#diff-fbb0617939d83339765219b2ac33504e3717bb1ae70d84801a8f3305612892adR82) [[2]](diffhunk://#diff-fbb0617939d83339765219b2ac33504e3717bb1ae70d84801a8f3305612892adR121)

**Dependency and import updates:**

* Added `once_cell` and updated `orb-info` dependencies in `ui/Cargo.toml` to support the new static initialization.
* Updated imports in `ui/src/engine/mod.rs` and `ui/src/main.rs` to include `orb_info::orb_os_release::{OrbOsRelease, OrbRelease}` and related types. [[1]](diffhunk://#diff-0478ce95268d2ac8c24e87b4ab6b88d48b03ed23470ffdc9b0abeef3cab048e9R5-R6) [[2]](diffhunk://#diff-e108e69cf0f5933cb78e6fed28b1c742b6ec02401d348c79d689db091b3aefcbR4-R6)


ui/src/engine/mod.rs
```diff
@@ -34,6 +34,22 @@ pub enum OrbType {
 
 pub const LED_ENGINE_FPS: u64 = 30;
 
+pub(crate) fn pearl_operator_default() -> Argb {
+    if *crate::RELEASE_TYPE == "analysis" {
```
AlexKaravaev: this is not good, why can't you reuse existing type?
`OrbRelease::Analysis => Argb::PEARL_OPERATOR_ANALYSIS`

ui/src/main.rs
```diff
@@ -85,6 +87,15 @@ struct BeaconArgs {
     duration: Duration,
 }
 
+fn current_release_type() -> Result<String> {
+    let os_release = OrbOsRelease::read_blocking().wrap_err("failed reading /etc/os-release")?;
```
AlexKaravaev: we shouldn't use read_blockin here probably?

ui/src/main.rs
```diff
@@ -1,6 +1,8 @@
 #![forbid(unsafe_code)]
 
 use humantime::parse_duration;
+use once_cell::sync::Lazy;
```
AlexKaravaev: we can reuse OnceLock

ui/src/main.rs
```diff
@@ -85,6 +87,15 @@ struct BeaconArgs {
...
 }
 
+fn current_release_type() -> Result<String> {
+    let os_release = OrbOsRelease::read_blocking().wrap_err("failed reading /etc/os-release")?;
+
+    Ok(os_release.release_type.as_str().to_owned())
+}
+
+pub(crate) static RELEASE_TYPE: Lazy<String> =
+    Lazy::new(|| current_release_type().unwrap());
```
AlexKaravaev: this would panic, we should default to non-analysis release type

AlexKaravaev: Why do we need global static variable here?

AlexKaravaev (changes requested):

ui/src/main.rs
chrisgalanis: I took the idea from: `priv-orb-core` 
```
fn current_release_type() -> Result<String> {
    let os_release = OrbOsRelease::read_blocking().wrap_err("failed reading /etc/os-release")?;
    Ok(os_release.release_type.as_str().to_owned())
}

```

ui/src/main.rs
AlexKaravaev: @chrisgalanis 

1. Please resolve comments once they are solved
2. This comment is still not resolved right?

chrisgalanis: 1. Thanks for the review ❤️ 
2. I think now it makes sense why it is static, with new code

AlexKaravaev (approved):

PR #1164, opened 2026-04-21
feat(zenorb): zoci ext methods

vmenge:
## changes
adds and documents extension methods for the Zenoh Orb Command Interface: small helpers for sending commands using zenoh queries and using strings/json underneath

small example, using utility methods:
- `Query::args`
- `Query::res`
- `zenorb::Session::command_raw`
- `Result<Reply, ReplyErr>::json`

```rust
use zenorb::zoci::{ReplyExt, ZociQueryExt};

blue.receiver(())
    .queryable("tuple", async |_ctx, query| {
        let args: (String, String) = query.args()?;
        query.res(&args).await?;

        Ok(())
    })
    .run()
    .await?;

let red = Zenorb::from_cfg(client_cfg.clone())
    .orb_id(orb_id.clone())
    .with_name("red")
    .await?;

let reply = red.command_raw("blue/tuple", "one two").await?;
let reply: Result<(String, String), String> = reply.json()?;
```


zenorb/src/sender.rs
```diff
@@ -30,6 +32,75 @@ impl Sender {
...
+    ///     },
+    /// ).await?;
+    ///
+    /// let reply: Result<StatusResponse, StatusError> = reply.json()?;
+    /// ```
+    pub async fn command(
+        &self,
+        keyexpr: &str,
+        payload: impl Serialize,
+    ) -> Result<Result<Sample, ReplyError>> {
```
sky-mart, line 60: 2 results?

vmenge: yeee

1st one is infra error
2nd one must be handled, `ReplyErr` contains actual error payload which is nice from zenoh

look a little awkward on type but usage in practice is pretty smooth bc outer error is `color_eyre` (see tests / examples)

sky-mart:
I'm not yet familiar with all rust features used, but overall looks good. 
I don't get 2 things:
1) Where is the command finally executed?
2) Can I create my util using this crate to connect to an orb's zenoh and run commands on an orb?

sky-mart (approved):

vmenge:
> I'm not yet familiar with all rust features used, but overall looks good. I don't get 2 things:
> 
> 1. Where is the command finally executed?
> 2. Can I create my util using this crate to connect to an orb's zenoh and run commands on an orb?

1) if you declare it in a `zenorb::Receiver`, its guaranteed to run in a `tokio::task` in the background once run is called ([example](https://github.com/worldcoin/orb-software/blob/5bc9a71326f169da8ebbe6ab6ac835310c1b76b5/zenorb/tests/zoci.rs?plain=1#L85))

2) yeee 100% :D  (tho only dev orbs)

PR #1167, opened 2026-04-23
feat(se050): certificate validation

TheButlah:
Implements validation of the SE050 certificate file, validating the cert chain starting from the pinned root NXP certificate.

The only ai-assisted portion of this pr is the contents of `se050/src/certs/null_params_alg.rs`, and `se050/example_data/2A66F1B2_bad_signature.crt`

Strictly speaking, actually validating the chip unique cert against the NXP root of trust isn't necessary since the assumption is that the backend pins the die-unique cert for a given orb already. But my hope is that the additional checks against the root cert further harden the system.

Some limitations (which I think are fine):
* we dont check CRLs when validating the cert chain. This is because doing so would require networking and greatly complicate the implementation.


mriise (approved):
I think it would be good to document and/or name the nxp certs to something more understandable at first glance, but otherwise-

LGTM

mriise:
🎉

se050/src/certs/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+        let end = ROOT_RANGE
+            .1
+            .min(INTERMEDIATE_RANGE.1)
+            .min(END_ENTITY_RANGE.1);
+
+        (start, end)
+    }
+
+    #[test]
+    fn test_known_cert_not_expired() {
```
BulatAbuzarov, line 155: not clear what this test actually for: 
the dates hardcoded from certs will be always correct for this until certs are replaced.

compute_total_range -> which is actually intersected range should be renamed to compute_total_valid_range then and would be sufficient.

as all other values already in this range:
```
            END_ENTITY_RANGE.0,
            INTERMEDIATE_RANGE.1,
            start + Duration::from_secs(1),
            end - Duration::from_secs(1),
```

The only way to know that cert is not expired is to somehow get actual current time, but that would reveal issue in future that intermediate certificate will be expired

TheButlah: The test is to check that the code that verifies cert expiry actually works. This test feeds it a number of different dates that are known to not be expired. A separate test checks that another set of dates that are known-to-be-expired are indeed rejected as expired.

These tests are important because without them, we have no evidence that the certificate verification functionality is actually validating dates properly

TheButlah:
aligned with @BulatAbuzarov, he approved (in dms)

se050/src/certs/mod.rs
TheButlah: aligned in dms

BulatAbuzarov (approved):

PR #1171, opened 2026-04-23
chore(ci): added job for ai labeling

TheButlah:
For now, the job is not required to pass. Its there to simply remind people to label their prs. Later I would like to make it required, but I'd wait to get opinions from others on the team for that (and would do that at a later date).

### DIFF
--- a/.github/workflows/pr-compliance.yaml
+++ b/.github/workflows/pr-compliance.yaml
@@ -7,6 +7,8 @@ on:
       - edited
       - synchronize
       - reopened
+      - labeled
+      - unlabeled
 
 permissions:
   pull-requests: read
@@ -35,4 +37,19 @@ jobs:
         run: |
           echo "❌ Please replace the PR template placeholder with a real description."
           exit 1
+
+  require-ai-label:
+    name: Require AI Label
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for ai label
+        env:
+          LABELS: ${{ toJson(github.event.pull_request.labels) }}
+        run: |
+          if echo "$LABELS" | jq -e 'map(.name) | any(startswith("ai:"))' > /dev/null 2>&1; then
+            echo "✅ Found an ai: label."
+          else
+            echo "❌ PRs must be labeled according to their ai use. Please label the PR."
+            exit 1
+          fi
 


pophilpo:
I am not against this job, and not against having to labels PRs (although, I am guilty of forgeting to put the label, so this PR would fix that :D ). But I struggle to get the meaning/implication of having this label overall
Is that for the reviewer to treat the PR differently ? In my mind, AI/non-AI PRs should get the same review. 

Labels might even bring worse reviews (maybe)
e.g. The reviewer might be annoyed to look at a fully vibe coded PR, so he just merges (or skims through code very quick) because the tests "pass"
Or, the reviewer might see that the PR was fully written by hand by a very respected high quality programmer :D and give the PR more trust from the get go, again failing to force himself to go through code thoroughly. But that might be even the case without labels, hard to tell honestly

TheButlah:
The old adage is "trust but verify" - which doesn't work here at all. The problem is that we cannot actually trust the code by default because we don't know if it is how it is because it was written by hand or if it was intentional. And often we are too busy to verify (most prs have always gotten a lgtm and very rarely do people actually review).

Labeling the PR is an important signal to help the reviewer understand the level of rigor with which to inspect the code, and a signal to not just say "oh this bit is a bit odd but I'm sure Bob had a good reason for doing it that way" when in reality it was odd because it didn't make sense - no one was at the wheel

Also it's just interesting data, and it helps inform whether it's reasonable to be nitpicky. If it's a vibe coded pr, there's no reason not to push on the author to make changes, but if it's hand written in a particular way the author did it that way intentionally and it's annoying to them to be opinionated. I don't mind pestering Claude to rewrite the code according to some approach, but it works differently when it's a human writing it with line-by-line intent

sfikastheo:
> AI/non-AI PRs should get the same review

I agree and this was also the same [expectation](https://tfh.slack.com/archives/C052M1G8SMA/p1773753296975139?thread_ts=1773673141.860359&cid=C052M1G8SMA) by @AlexKaravaev .
If you push shit, either because your code is shit or your clanker pushed it, does not matter.
> but it works differently when it's a human writing it with line-by-line intent

I agree with the sentiment, but not the implementation. I see labels for this just as further bureaucracy.
Thus far, i have been using good old judgement and we should have no fear pushing back. I find it easy **still** to tell when its ai-slop or when there was intent behind it. Worse case you try correcting something and someone can explain their reasoning behind it, if there was one. It is also informative for the reviewer since he is getting perspective on something that feels off.

We operate with accountability & tags like proposed i have a feeling that they just hide that accountability by -> "oh its ok MyAgentsName did it, its good enough it works its vibed take this up with claude".

If you ever  feel there is an issue with the reviewing culture in the company/team, we might as well tackle the issue properly with a zero-bullshit culture with a discussion around it.
 
Almost everyone is using AI models, and if you are not, you should so that you understand better the possibilities of them, where they can be useful and where they fall short.

vmenge:
honestly i don't like the idea :sweat_smile: 
i think we should review every code the same. we've had problems before with thousands of line of slop, yes, but it has gone down and i think we should simply put the hammer down on holding people accountable for their contributions. 

adding an ai label will just make the review get different treatment, and it really shouldn't imo.

mistakes can be made with vibe coding, heck, i've had my own occurrence of being in a hurry and accidentally submitting slop in a pr. it was a small change, the pr was reviewed by a regular reviewer who caught the slop and i adjusted it.

if external contributors start submitting thousands of lines of slop and using us as the reviewer of the clanker's code, then we reach out and ask them to properly review the code or else we will stop accepting their submissions. internally i think this is a non-issue and everyone is on board. 

mistakes will happen but we are all on the same page: you push it, you own it. same it always has been 

so my 2c is we shouldn't push for this label. if someone wants to label their ai code, i would say go for it, but i dont think we should be actively encouraging people to do it, or giving them warnings because they didn't do it

TheButlah:
closing until further discussion

PR #1173, opened 2026-04-24
chore(hil): add ssh config for the orbs

sky-mart:
to do `ssh orb` or `scp something orb:~/` from the hil


sfikastheo (approved):
cool  :)

PR #1198, opened 2026-05-10
feat(hardening): per-service least-privilege users

pophilpo:
# Motivation

  All Worldcoin Orb services previously ran as the worldcoin user, which has root-equivalent privileges on the system. This PR replaces that model with dedicated, least-privilege system accounts for each service. Each account receives only the Linux capabilities, filesystem access, and supplementary groups required for its specific function — nothing more.

Four services are hardened in this PR. Other services will be addressed in a separate workstream.

# Services hardened

  ## orb-attest

  Changes:
  - Runs as dedicated orb-attest user/group
  - Added to worldcoin-dbus group for session bus socket access (/tmp/worldcoin_bus_socket)
  - Ambient + bounding capability set reduced to CAP_NET_RAW only (required by orb-mcu-util for raw CAN-FD socket to reboot security MCU)
  - Device access restricted: only `/dev/i2c-*`  and char-tee
  - Address families restricted to `AF_INET AF_INET6 AF_UNIX AF_CAN`

  ## orb-backend-status

  Changes:
  - Runs as dedicated orb-backend-status user/group
  - Added to worldcoin-dbus group for session bus socket access
  - Zero capabilities (AmbientCapabilities=, CapabilityBoundingSet=) — this service needs none
  - PrivateDevices=yes (no device access required)
  - Address families restricted to `AF_INET AF_INET6 AF_UNIX`

  ## orb-connd

  Changes:
  - Runs as dedicated orb-connd user/group
  - Added to netdev group: wpa_cli accesses the wpa_supplicant control socket at /var/run/wpa_supplicant/wlan0 (udev: GROUP=netdev MODE=0660)
  - Added to tee group: on Diamond, orb-connd spawns itself as the orb-ss-connd-nmprofiles subprocess for OP-TEE secure storage. The subprocess inherits supplementary groups across fork()+exec(), so the parent must hold the tee group membership.
  - PrivateDevices=no, DeviceAllow=char-tee rw for Diamond OP-TEE path
  - PrivateTmp=no — session D-Bus socket lives at /tmp/worldcoin_bus_socket; a private tmpfs would hide it

  ## orb-update-verifier

  Changes:
  - Previously ran as root; now runs as dedicated orb-update-verifier user/group
  - Added to tty group for orb-mcu-util UART access (/dev/ttyTHS*)
  - Ambient + bounding capabilities scoped to exactly what's needed:
    - CAP_NET_RAW — CAN-FD SOCK_RAW socket for orb-mcu-util
    - CAP_LINUX_IMMUTABLE — clear FS_IMMUTABLE_FL on EFI variable files
    - CAP_FOWNER — efivarfs FS_IOC_SETFLAGS ioctl requires passing the inode ownership check; since EFI vars are owned by root and we are not root, CAP_FOWNER is required alongside CAP_LINUX_IMMUTABLE
    - CAP_DAC_OVERRIDE — write to PMC sysfs nodes (/sys/devices/platform/*/c360000.pmc/rootfs_retry_count_*) currently owned by root. TODO: replace with a udev rule setting GROUP=orb-slot on those nodes.
  - ProtectKernelTunables=no — service must write to /sys/firmware/efi/efivars/ and PMC sysfs; these would be blocked by the strict /sys read-only bind mount
  - ReadWritePaths carved out for EFI vars and both known PMC sysfs paths
  - Address families restricted to AF_UNIX AF_CAN

# Testing
- Tested each service and their functionality
- Tested a mock signup
- Tested OTA from the hardened PR version to something else


orb-connd/debian/worldcoin-connd.service
```diff
@@ -9,17 +9,82 @@ Requires=zenohd.service
...
 
+# Diamond only: spawns itself (via std::env::current_exe()) as the `orb-ss-connd-nmprofiles`
+# user/group for OP-TEE secure storage. tokio Command::uid()/gid() calls setresuid()/
+# setresgid() in the child after fork() but before exec(); those syscalls require CAP_SETUID
+# and CAP_SETGID in the effective capability set. Both caps are in AmbientCapabilities so
+# they are inherited into the child's effective set across fork().
+# On Pearl, the subprocess path is never taken (ProfileStorage::NetworkManager used instead);
+# these capabilities are unused there.
+AmbientCapabilities=CAP_SETUID CAP_SETGID
+CapabilityBoundingSet=CAP_SETUID CAP_SETGID
```
BulatAbuzarov, line 37: missing?
```
DeviceAllow=char-tee rw
```
secure storage is not used then and fallback to TMP storage

orb-connd/debian/worldcoin-connd.service
```diff
@@ -9,17 +9,82 @@ Requires=zenohd.service
...
+
```
BulatAbuzarov, line 38: Supplementary groups tee is missed

vmenge (approved):

attest/debian/orb-attest.worldcoin-attest.service
```diff
@@ -17,5 +22,63 @@ Restart=on-failure
...
+
+# If at some point this service will need write access make sure to change this
+ReadOnlyPaths=/
+# D-Bus socket lives at /tmp/worldcoin_bus_socket; PrivateTmp would hide it
+PrivateTmp=no
+# Deny access to all devices by default; explicitly allow only what is needed
+DevicePolicy=closed
+# orb-sign-attestation accesses SE050 via /dev/i2c-* and OP-TEE via /dev/tee*
+DeviceAllow=/dev/i2c-* rw
+DeviceAllow=char-tee rw
```
BulatAbuzarov: DeviceAllow=/dev/tee0 rw
otherwise  access give to teepriv0 that belongs to tee-supplicant too
```
udevadm info -a -n /dev/teepr 

Udevadm info starts with the device specified by the devpath and then
walks up the chain of parent devices. It prints for every device
found, all possible attributes in the udev rules key format.
A rule to match, can be composed by the attributes of the device
and the attributes from one single parent device.

  looking at device '/devices/virtual/tee/teepriv0':
    KERNEL=="teepriv0"
    SUBSYSTEM=="tee"
    DRIVER==""
    ATTR{implementation_id}=="1"
    ATTR{power/control}=="auto"
    ATTR{power/runtime_active_time}=="0"
    ATTR{power/runtime_status}=="unsupported"
```

pophilpo: Done 👍🏻

BulatAbuzarov (approved):

attest/debian/orb-attest.worldcoin-attest.service
```diff
@@ -1,14 +1,19 @@
...
+After=worldcoin-dbus.service
 After=worldcoin-dbus.socket
 Wants=worldcoin-dbus.socket
 
 [Service]
 Type=simple
-User=worldcoin
+User=orb-attest
+Group=orb-attest
+SupplementaryGroups=worldcoin-dbus
```
BulatAbuzarov: SupplementaryGroups=tee ?

BulatAbuzarov (approved):
LGTM!

PR #1209, opened 2026-05-13
feat: try_lock/unlock hil commands

sky-mart:
"


vmenge (approved):

nix/scripts/hil_try_lock.py
```diff
@@ -29,19 +33,6 @@ def main() -> int:
     orb_id = os.environ["HIL_ORB_ID"]
     note = " ".join(sys.argv[1:]).strip() or None
 
-    runners = requests.get(f"{orchestrator}/runners", timeout=TIMEOUT).json()
```
chrisgalanis: This should be handled by the `hil-server` that you are making the request and not the client.

PR #1213, opened 2026-05-15
feat: handle gondor in orb-supervisor via zoci

sky-mart:
In HIL tests we have to run gondor with sudo. Reusing the zoci machinery would solve this necessity.

In this PR I move the job handler from orb-jobs-agent to orb-supervisor.

Fleet Commander Job: https://www-fleet-http.internal.stage.orb.worldcoin.dev/jobs/2f9003aa-fa1f-419b-8f16-24559ac267b3?tz=Europe%2FBerlin

`zorb -o bce8243c -r 10.42.0.206 query supervisor/job/gondor "to-latest"` is also working.


vmenge (approved):
lgtm!

PR #1226, opened 2026-05-22
feat(attest): SE050 key migration selection at startup

BulatAbuzarov:
On startup, check whether `worldcoin-se050-provision.service` is active (via systemd D-Bus) to decide which key set to use. If migrated keys are present, wait for the backend to become reachable, then attempt a token round-trip with the migrated key. If the backend does not yet have the key registered, submit an NXP-attested proof and poll for up to 60 s until the backend accepts the migrated key.

Expose the result as `new_keys_active` on the `AuthTokenManager1` D-Bus interface

Added feature flag to skip migration phase from current release until issues are resolved
```
2026-06-16T20:01:37.160597Z  INFO orb_attest: worldcoin-se050-provision.service not active, using legacy keys
```

with feature flag on test orb with migrated keys:
```
2026-06-16T20:35:50.174936Z  INFO orb_attest: backend already accepts migrated keys

```

### DIFF
--- a/attest/src/lib.rs
+++ b/attest/src/lib.rs
@@ -92,20 +92,26 @@ pub async fn main() -> eyre::Result<()> {
 
     let (is_online_tx, is_online_rx) = watch::channel(false);
 
-    let zenorb = Zenorb::from_cfg(zenorb::default_cfg())
+    match Zenorb::from_cfg(zenorb::default_cfg())
         .orb_id(orb_id.clone())
         .with_name("attest")
-        .await?;
-
-    let _ = zenorb
-        .receiver(is_online_tx)
-        .querying_subscriber(
-            "connd/oes/active_connections",
-            Duration::from_secs(1),
-            update_is_online,
-        )
-        .run()
-        .await?;
+        .await
+    {
+        Ok(zenorb) => {
+            let _ = zenorb
+                .receiver(is_online_tx)
+                .querying_subscriber(
+                    "connd/oes/active_connections",
+                    Duration::from_secs(1),
+                    update_is_online,
+                )
+                .run()
+                .await?;
+        }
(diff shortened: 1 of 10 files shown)


vmenge (approved):
LGTM 🚀

attest/src/lib.rs
```diff
@@ -89,6 +130,167 @@ pub async fn main() -> eyre::Result<()> {
...
+        }
+        sleep(KEY_ACTIVATION_POLL_INTERVAL).await;
+    }
+
+    warn!(
+        "backend did not accept migrated keys after {}s; proceeding with legacy keys",
+        KEY_ACTIVATION_POLL_TIMEOUT.as_secs()
+    );
+    false
+}
```
pophilpo, line 275: A nit, but can happen with our backend :) 

If the backend is slow, the polling loop may only fire a couple of times instead of the intended ~12. The deadline is checked after each request returns, not before, so a single slow request can eat most of the 60s window, resulting into fewer tries.

The fix would be either to make HTTP requests with a timeout, or wrap this whole thing into a tokio timeout, that would drop a request mid point if it takes too long. Something like this idk

```
let result = tokio::time::timeout(KEY_ACTIVATION_POLL_TIMEOUT, async {
    loop {
        if remote_api::try_token_with_migrated_key(orb_id, auth_url).await {
            return true;
        }
        sleep(KEY_ACTIVATION_POLL_INTERVAL).await;
    }
})
.await;

match result {
    Ok(true) => {
        info!("backend now accepts migrated keys");
        true
    }
    _ => {
        warn!(
            "backend did not accept migrated keys after {}s; proceeding with legacy keys",
            KEY_ACTIVATION_POLL_TIMEOUT.as_secs()
        );
        false
    }
}
```

PR #1229, opened 2026-05-25
feat(orb-dogd): standardized crate for pushing metrics to datadog

sfikastheo:
after following @vmenge commands semi-religiously, this might do?

Main points:
- fire and forgets mentality on metrics
- default implementation spins a thread hosting the interaction between dogstatsd <-> process. 
- [testing] feature hosts implementations for mocking the MetricsEmitter & for testing the process logic from a recorder

### DIFF
--- a/orb-dogd/src/dd.rs
+++ b/orb-dogd/src/dd.rs
@@ -1,0 +1,63 @@
+use dogstatsd::{Client, DogstatsdError, Options};
+use flume::{Sender, TrySendError};
+use std::thread;
+use tracing::error;
+
+use super::{Metric, MetricEmitter, MetricError};
+
+pub struct DogstatsdClient {
+    tx: Sender<Metric>,
+}
+
+impl DogstatsdClient {
+    /// Bind the UDP socket and spawn the consumer thread,
+    /// propagating metrics to the metric collector
+    ///
+    /// Fails if [`Client::new`] fails (socket bind / permission).
+    /// The worker runs until all [`Sender`]s are dropped.
+    pub fn new() -> Result<Self, DogstatsdError> {
+        let (tx, rx) = flume::bounded(1024);
+
+        let datadog_options = Options::default();
+        let client = Client::new(datadog_options)?;
+
+        thread::spawn(move || {
+            while let Ok(metric) = rx.recv() {
+                match metric {
+                    Metric::Count { stat, val, tags } => {
+                        if let Err(e) = client.count(stat, val, tags) {
+                            error!("emitting metric failed with: {e}");
+                        }
+                    }
+                    Metric::Gauge { stat, val, tags } => {
+                        if let Err(e) = client.gauge(stat, val.to_string(), tags) {
+                            error!("emitting metric failed with: {e}");
+                        }
+                    }
+                    Metric::Histogram { stat, val, tags } => {
+                        if let Err(e) = client.histogram(stat, val.to_string(), tags) {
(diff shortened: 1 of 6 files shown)


orb-dogd/src/dd.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        let datadog_options = Options::default();
+        let client = Client::new(datadog_options)?;
+
+        thread::spawn(move || {
+            while let Ok(metric) = rx.recv() {
+                match metric {
+                    Metric::Count { stat, val, tags } => {
+                        if let Err(e) = client.count(stat, val, tags) {
+                            error!("emitting metric failed with: {e}");
```
sky-mart: how about 4 slightly different error messages?

orb-dogd/src/dd.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct DogstatsdClient {
+    tx: Sender<Metric>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum Metric {
+    /// Counter delta; Used by:
+    /// [`MetricEmitter::count`] and [`MetricEmitter::incr`]
+    Count {
+        stat: String,
```
sky-mart, line 17: what is stat here? is it a metrics name?

sky-mart (approved):

orb-dogd/src/dd.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Connect to the local statsd collector.
+    ///
+    /// Fails if the underlying socket cannot be bound.
+    pub fn new() -> Result<Self, DogstatsdError> {
+        let (tx, rx) = flume::unbounded();
+
+        let datadog_options = Options::default();
+        let client = Client::new(datadog_options)?;
+
+        thread::spawn(move || {
```
Qbicz, line 56: Hey! Here process like update-agent may emit a metric right before returning or exiting. With this implementation, the metric goes to a queue. If the process exits, the worker thread is killed with the process and Datadog may never receive the metric

orb-dogd/src/dd.rs
sfikastheo: Yes

orb-dogd/src/dd.rs
sfikastheo: the {e}, will be printed alongside the message, and each time its different

orb-dogd/src/dd.rs
sfikastheo: I agree, its a correct limitation 😮‍💨 .
It was discussed with @vmenge, and he did not think that a sync implementation was important here, but it is possible to create within the abstraction crate.

sfikastheo: The mental model was a "best effort, fire and forget API". 
I would need `async` to remove the limitation, and that forces each consumer to spin up a runtime.
Maybe a valid thing to be doing? - but i pushed back on this, i do believe that the lib should offer an implementation that is runtime-free.

vmenge: This is actually also a problem with logs as well, usually when updating some logs are lost.

We could add something to the DD client to allow waiting for all pending messages to resolve, but I would treat it as non-blocking :+1: 

We should as a follow up add graceful termination on `update-agent` handling sigterm to properly flush metrics and logs later on

vmenge (approved):
LGTM! Nice job :D

PR #1230, opened 2026-05-25
chore(update-agent): swap update-agent to dogd

sfikastheo:
moving the datadog metrics propagation to `orb-dogd`.

commit changelog:
1) minimal change -> sticking to the original static datadog client architecture with using `orb-dogd`
2) removing global state of the metrics-emitter instance. + fixing `map_err` metrics. The metric emittion was in the constructing block for the map_err closure, not in the closure itself. That means that it must have been firing regardless of an error or not.
3) adds `orb.platform.update.failed` metric if the target's slot status already is `UPD_IN_PROG`, fixes ORBS-1469
    adds `orb.platform.update.fallback` metric if the target's slot status is `UNBOOTABLE`, fixes ORBS-1470


update-agent/src/main.rs
```diff
@@ -331,6 +335,25 @@ fn run(args: &Args) -> eyre::Result<()> {
...
 
+    // check the status of the target slot to detect prior update attempts:
+    // - `UpdateInProcess` -> consumption was interrupted (shutdown, crash,
+    //   or boot failed before L4TLauncher cleared the state).
+    // - `Unbootable` -> the bootloader exhausted retries on the target slot
+    //   and fell back to the currently active slot.
+    match slot_ctrl.get_rootfs_status(target_slot.into()) {
+        Ok(orb_slot_ctrl::RootFsStatus::UpdateInProcess) => {
+            let _ = metrics
+                .incr("orb.platform.update.failed", orb_dogd::NO_TAGS)
```
Qbicz, line 346: Here the same comment applies as in https://github.com/worldcoin/orb-software/pull/1229#discussion_r3302604546

update-agent/src/update/can.rs
```diff
@@ -174,11 +172,15 @@ impl Update for components::Can {
...
                 })
-                .map_err({
-                    DATADOG
-                        .incr("orb.update.count.component.can", ["status:write_error"])
-                        .or_log();
-                    |e| e
+                .inspect_err(|_| {
+                    let _ = metrics
+                        .incr(
+                            "orb.platform.update.component.can",
```
sky-mart: can be a named constant

update-agent/src/update/gpt.rs
```diff
@@ -177,9 +178,12 @@ impl Update for components::Gpt {
             .wrap_err_with(|| format!("GPT disk `{}` flush failed", self.device))?;
         debug!("-- flushed!");
 
-        DATADOG
-            .incr("orb.update.count.component.gpt", ["status:write_complete"])
-            .or_log();
+        let _ = metrics
+            .incr(
+                "orb.platform.update.component.gpt",
+                ["status:write_complete"],
```
sky-mart: statuses are also repeated afais, how about an enum with `as_str`?

sky-mart (approved):

update-agent/src/main.rs
sfikastheo: there is nothing killing the agent here, but yea the worry is absolutely correct. 
Maybe the idea is not correct and we should go back to the `static_ Lazy::new()` client... 
I am really conflicted.

update-agent/src/update/can.rs
sfikastheo: done

vmenge (approved):

PR #1231, opened 2026-05-26
chore(dogd): use a socket; socket retry logic

vmenge:
## changes
- uses a socket as default for `DogstatsdClient`
- adds retry logic for client creation and makes `DogstatsdClient::new` infallible

### DIFF
--- a/orb-dogd/src/dd.rs
+++ b/orb-dogd/src/dd.rs
@@ -1,13 +1,16 @@
 use dogstatsd::{Client, DogstatsdError, Options};
 use flume::Sender;
-use std::thread;
-use tracing::warn;
+use std::{fs, path::Path, thread};
+use tracing::{error, info, warn};
 
 use super::{MetricEmitter, MetricError};
 
 pub struct DogstatsdClient {
     tx: Sender<Metric>,
 }
+
+const DOGSTATSD_SOCKET_PATH: &str = "/run/datadog/dsd.socket";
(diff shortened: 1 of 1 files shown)


orb-dogd/src/dd.rs
```diff
@@ -43,17 +46,45 @@ pub(crate) enum Metric {
...
     /// Fails if the underlying socket cannot be bound.
-    pub fn new() -> Result<Self, DogstatsdError> {
+    pub fn new() -> Self {
         let (tx, rx) = flume::unbounded();
 
-        let datadog_options = Options::default();
-        let client = Client::new(datadog_options)?;
-
         thread::spawn(move || {
+            let client = loop {
```
sky-mart, line 63: are we waiting forever in the worst case?

sky-mart (approved):

orb-dogd/src/dd.rs
vmenge: pretty much! both `connd` and `jobs-agent` already do the same
alternatively we can bubble up the error, but then we either move forward without any metrics or we bubble up further causing us to either have similar logic on the caller, or by having systemd restart the applicaton

only problem here is the backing channel is unbounded. realistically should never happen, but if it does can cause a memory leak if consumers send too many messages before the DD socket either shows up or the client succeeds in connecting

im talking with @sfikastheo in a bit, we might bring back the bounded channels here and use `try_send` on the client or just leave it be

sfikastheo (approved):

PR #1238, opened 2026-05-28
refactor(connd): use new datadog client

vmenge:
## changes
- connd to use shared datadog client


sky-mart (approved):

vmenge:
> Found two issues:
> 
> 1. `orb-connd/src/main.rs:116` switches `connd` to `orb_dogd::DogstatsdClient`, but that client only waits for `/run/datadog/dsd.socket` and no longer preserves the old UDP fallback. The deleted local client explicitly fell back to `dogstatsd::Options::default()` when the socket was absent, so on systems where DogStatsD is reachable over UDP but the UDS is missing, `connd` will stop emitting metrics.
> 2. `orb-connd/src/reporters/datadog.rs:69`, `:76`, `:94`, `:100` and `orb-connd/src/reporters/data_usage.rs:47`, `:55` now discard `MetricEmitter` errors. `orb-dogd` returns errors when its bounded queue is full or the worker dies; the old code propagated these with `?`, so the reporter logged/restarted. Please at least log these failures, otherwise metric loss from backpressure/worker failure becomes invisible.
> 
> I couldn’t run `cargo check` in this sandbox because rustup failed to create temp files on the read-only filesystem.

two issues according to whom 

plz evaporate

PR #1240, opened 2026-05-29
feat: support forking of the dogd client

sky-mart:
"


sky-mart:
useless, because we always use fork+exec

PR #1241, opened 2026-05-29
feat(gondor): cleanup gondor invocations

sfikastheo:
* `supervisor`: instead of wrapping another python script, sanitizes the input, and exports it to systemd
* `orb-hil`: instead of calling gondor, use `zorb` from the runner to query supervisor's job
* `update-agent`: consume `version_overwrite`, if that is provided, overwrite `/etc/os-release` with that value.

The `gondor` binary will be removed from `development` builds, and be replaced by an alias to `zorb`querying 
`supervisor`.

This is a development feature to quickly allow developers to OTA to whatever they wish.
PR is spread upon commits touching the different crates separately

### DIFF
--- a/hil/src/commands/ota/system.rs
+++ b/hil/src/commands/ota/system.rs
@@ -4,8 +4,6 @@ use color_eyre::{
     Result,
 };
 
-const GONDOR_CALLS_FOR_OTA_PATH: &str = "/usr/local/bin/gondor-calls-for-ota";
-
 /// Reboot the Orb device using orb-mcu-util and shutdown
 pub async fn reboot_orb(session: &RemoteSession) -> Result<()> {
     session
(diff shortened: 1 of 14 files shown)


hil/src/commands/ota/system.rs
```diff
@@ -66,42 +64,37 @@ fn parse_slot_from_output(output: &str) -> Result<String> {
...
-        format!("TERM=dumb sudo {GONDOR_CALLS_FOR_OTA_PATH} '{escaped_target}'");
-    let result = session
-        .execute_command(&command)
+    let payload = serde_json::json!({
+        "version": target_version.trim(),
+        "restart": true,
+    })
+    .to_string();
+
+    let output = tokio::process::Command::new("zorb")
```
sky-mart, line 82: how about using the lib here? I don't mind against zorb, but it's 2 rust binaries

sfikastheo: Yep, can do. seemed more invasive, did not know if we want to bake in zenoh & it was already calling another binary. 
Additionally IMO, we should try to swap most of all of this into python anyway, It already works and the interactive shell makes it actually beautiful to work with. 

Seems like an attempt at properly patching dead code

supervisor/src/tasks/zoci.rs
```diff
@@ -1,77 +1,168 @@
...
-        }
-        let output = cmd.output().await.wrap_err("failed to spawn gondor")?;
+async fn export_update_version(
+    conn: &zbus::Connection,
+    var: &str,
+    value: &str,
+) -> Result<()> {
+    let manager = ManagerProxy::new(conn).await?;
+    manager
+        .set_environment(vec![format!("{var}={value}")])
```
sky-mart, line 59: No bind-mounting of /etc/os-release is required anymore?

sfikastheo: nop 🏖️ 💘 ⭐ 🌟

sky-mart (approved):
Current HIL tests still send the `no_restart` option, so when you update the binaries in orb-os, the OTA test is supposed to fail.

LGTM

vmenge (approved):

PR #1248, opened 2026-06-03
feat(security-utils): add test for certificate expiration

sfikastheo:
adding a test to validate pinned certificate expirations

### DIFF
--- a/security-utils/src/reqwest.rs
+++ b/security-utils/src/reqwest.rs
@@ -213,6 +213,9 @@ mod tests {
             .expect("Failed to make GTS R4 cert");
     }
 
+    // assert expiration date is under 6 months
+    const EXPIRATION_GATE: Duration = Duration::from_secs(180 * 24 * 60 * 60);
+
     #[test]
     fn pinned_certificates_do_not_expire_soon() {
         for (name, cert_pem) in [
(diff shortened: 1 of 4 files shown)


security-utils/src/reqwest.rs
```diff
@@ -210,4 +212,45 @@ mod tests {
...
+        ] {
+            let (_, pem) = x509_parser::pem::parse_x509_pem(cert_pem)
+                .unwrap_or_else(|e| panic!("{name} should parse as PEM: {e}"));
+            let cert = pem
+                .parse_x509()
+                .unwrap_or_else(|e| panic!("{name} should parse as X.509: {e}"));
+
+            // assert expiration date is under 6 months
+            const EXPIRATION_GATE: Duration = Duration::from_secs(180 * 24 * 60 * 60);
+
```
pophilpo: Do this outside the loop maybe? Just feels werid.
Although I am being very nitpicky since this doesn't really matter

pophilpo (approved):

security-utils/src/reqwest.rs
sfikastheo: ah yea you are right missed it, patching it

PR #1250, opened 2026-06-04
feat(backend-status): add orb stand QR type to CoreStats

danielle-tfh:
## Motivation
The FM backend already receives the orb stand QR id, but can't tell whether a scan came from a tablet screen or a static printed sign.

Adds `orb_stand_qr_type` ("tablet"/"static") and `orb_stand_qr_utm_source` to `CoreStats` and threads both through to the `OrbStatusApiV2` payload sent to the backend. orb-core will populate them in a follow-up.


AlexKaravaev (approved):
LGTM but I think you also need to add new field in status request on backend side in `orb-fleet-backend` for it to be received/processed

PR #1251, opened 2026-06-05
feat(attest): success ratio, duration and error cause metrics

vmenge:
## changes
adds distribution and count metrics for duration, success ratio and error causes

## notes
we only count the metrics when we have a stable connection , otherwise delays or error causes might be just because the orb is simply not connected to the internet


vmenge:
> Found one concrete issue:
> 
> * `attest/src/lib.rs:213` / `attest/src/lib.rs:218-219`: `ConnectivityTracker` never marks the connection as unstable. It initializes `stable_connection` to `true`, ignores the receiver’s current value, and when an offline update arrives it stores `true` again. As a result, token refresh metrics are still emitted when the Orb starts the attempt already offline or loses connectivity during the attempt, which defeats the PR’s stated filtering behavior. This should likely initialize from `is_online_rx.borrow()` and store `false` when `!*is_online_rx.borrow()`, with a test covering both “already offline” and “disconnect during refresh”.

good clanker

vmenge:
> Found two concrete issues:
> 
> * `attest/src/lib.rs:51`, `attest/src/lib.rs:212`: `ConnectivityTracker` starts with `stable_connection = true` and only reacts to future `watch::changed()` events. Since the channel is initialized to `false`, token refresh attempts made before the first connd sample, or while the current value is already offline, still emit metrics as if connectivity was stable. This defeats the PR’s “only count metrics when we have a stable connection” behavior. Initialize from `*is_online_rx.borrow()` and treat an initial `false` as unstable.
> * `attest/src/lib.rs:53`, `attest/debian/orb-attest.worldcoin-attest.service:1`: `orb-attest` now opens a Zenoh session during startup and exits if Zenoh is unavailable after retries, but the systemd unit only adds the `zenohd` group and does not add `Requires=zenohd.service` / `After=zenohd.service`. Other Zenoh consumers in this repo declare that dependency. Without it, attest can fail or restart-loop during boot ordering races, blocking token availability.

bro you dumb

attest/src/remote_api.rs
```diff
@@ -556,27 +561,99 @@ async fn get_token_inner(
...
+                }
+
                 return token;
             }
+
             Err(e) => {
                 let retry_delay = e.retry_delay(delay);
                 error!("failed to get token: {e}, retrying in {retry_delay:?}");
                 sleep(retry_delay).await;
                 delay = e.next_retry_delay(delay);
```
pophilpo: Vaugepost:
Fix.

vmenge: image
outdated by the time u posted it

image

pophilpo (approved):

vmenge:
> Found two concrete issues:
> 
> * `attest/debian/orb-attest.worldcoin-attest.service:16` / `attest/src/lib.rs:53`: `orb-attest` now opens a Zenoh client socket at startup, but the unit only adds the `zenohd` group and does not declare `Requires=zenohd.service` / `After=zenohd.service`. Other Zenoh consumers in this repo do. This creates a startup ordering regression where attest can fail before fetching any token if `zenohd` is not up yet. Either add the systemd dependency or make the metrics subscription non-fatal.
> * `attest/src/lib.rs:212`: `ConnectivityTracker::start` initializes `stable_connection` to `true` and only flips it on future `changed()` events. If the watch channel is already `false` when an attempt starts, including the initial value from `watch::channel(false)` or a cached offline OES sample, the attempt is still counted as having a stable connection. That defeats the PR’s stated metrics gate and will emit success/failure metrics while offline. Initialize from `*is_online_rx.borrow()` or otherwise require the current value to be online before emitting.

stop flooding my pr please

attest/src/lib.rs
```diff
@@ -25,22 +38,41 @@ pub const SYSLOG_IDENTIFIER: &str = "worldcoin-attest";
...
+    let zenorb = Zenorb::from_cfg(zenorb::default_cfg())
+        .orb_id(orb_id.clone())
+        .with_name("attest")
+        .await?;
+
+    let _ = zenorb
+        .receiver(is_online_tx)
+        .querying_subscriber(
+            "connd/oes/active_connections",
+            Duration::from_secs(1),
```
BulatAbuzarov, line 62: hardcoded number

BulatAbuzarov (approved):

PR #1257, opened 2026-06-08
feat(connd): primary conn metric, capability metric

vmenge:
## new metrics
- lte capability
- primary connection

## tested on
- [ ] diamond orb


vmenge:
not needed

PR #1261, opened 2026-06-15
feat(dogd): handle backpressure

vmenge:
## changes
introduces backpressure handling to avoid issue seen when burst emitting full queue of metrics
image

limits metrics to 25/50ms (500/s). this should avoid the errno 11 seen in the image above, most likely caused by multiple clients overwhelming the socket/agent.

## drawbacks
if we add any metrics where the _emitting itself_ (not the value of the metric) is time sensitive, this could cause issues if we have a hot loop. worst case scenario we deviate a bit on distribution of some metric time buckets up to 5s. if this becomes an issue we might prefer to drop sending these metrics

### DIFF
--- a/orb-dogd/src/dd.rs
+++ b/orb-dogd/src/dd.rs
@@ -15,6 +15,7 @@ const QUEUE_SIZE: usize = 1024;
 const MAX_EMIT_PER_TICK: usize = 20;
 const TICK: Duration = Duration::from_millis(100);
 
+#[derive(Clone)]
 pub struct DogstatsdClient {
     tx: Sender<Metric>,
 }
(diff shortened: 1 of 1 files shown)


vmenge:
> Found one concrete issue:
> 
> * `orb-dogd/src/dd.rs:15` / `orb-dogd/src/dd.rs:16`: the throttle is set to `25` emits per `50ms`, which is `500/s`, not the intended `200/s` described for avoiding DogStatsD socket backpressure. Since the mitigation depends on the lower send rate, this can still overload the agent/socket under bursts. Consider `20` per `100ms` or add a test asserting the effective rate.
> 
> I couldn’t run formatting/tests in this sandbox because rustup needs to write temp files under a read-only filesystem.

i just pushed the commit before i updated PR, leave me alooooooone

pophilpo (approved):
Very good 👍🏻

PR #1266, opened 2026-06-17
feat: pcp-messages protobuf

karelispanagiotis:
Introduces protobuf definitions for pcp messages to be included in the pcp-package.


pcp-messages/proto/pcp/v1/di_iris_embedding_shares.proto
```diff
@@ -0,0 +1,17 @@
...
+
+message DiIrisEmbeddingShares {
+  DiIrisEmbeddingShareV1 share_v1 = 1;
+}
+
+message DiIrisEmbeddingShareV1 {
+  string model_version = 1;
+  string shares_version = 2;
+
+  string left_share = 3;
```
gryaele: Can we make these byte arrays?

pcp-messages/proto/pcp/v1/di_iris_embeddings.proto
```diff
@@ -0,0 +1,23 @@
...
+message DiIrisEmbeddings {
+  DiIrisEmbeddingV1 embedding_v1 = 1;
+}
+
+message DiIrisEmbeddingV1 {
+  string model_version = 1;
+  string embedding_inference_backend = 2;
+  string embedding_version = 3;
+
+  string left_embedding = 4;
```
gryaele: These could be byte arrays

Cargo.toml
```diff
@@ -41,6 +41,7 @@ members = [
 	"orb-info",
 	"orb-jobs-agent",
 	"ota-backend",
+	"pcp-messages/rust",
```
andronat: should we name this pcp-defs? my problem with "messages" is that we are defining a format of the pcp. not a message protocol

pcp-messages/README.md
```diff
@@ -0,0 +1,8 @@
+# pcp-messages
+
+Protobuf definitions for PCP (Personal Custody Package) messages.
+
+- `proto/`   — protobuf source files
+- `buf.yaml` — Buf module config (lint + breaking-change rules)
+- `rust/`    — Rust crate `orb-pcp-messages`
+- `go/`      — Go bindings (placeholder, to be set up)
```
andronat: I think these lines can easily become stale and don't add much info. i think we should remove them

Cargo.toml
karelispanagiotis: no problem. I didn't know how to name it, I just chose this one because it includes the definitions of protobuf-messages

pcp-messages/proto/pcp/v1/di_iris_embeddings.proto
karelispanagiotis: we are now using `repeated sint32`

pcp-messages/proto/pcp/v1/di_iris_embedding_shares.proto
karelispanagiotis: we are now using `repeated uint32`

pcp-defs/go/README.md
```diff
@@ -0,0 +1,3 @@
+# go
+
+Placeholder for the Go bindings of `pcp-defs`.
```
leonanos8: Based on our internal discussion, we wont need go bindings.
We are going to treat signup-service(Go service) as a middleman that just forwards opaque bytes, so the only entity that will need parsing code is deep-identifier-ampc which is in Rust and can import with smth like:

```
[dependencies]
orb-pcp-defs = { git = "https://github.com/worldcoin/orb-software", rev = "<pin-a-commit>" }
```
and then the types will be available with orb_pcp_defs::prost::Message

leonanos8 (approved):

pcp-defs/go/README.md
karelispanagiotis: removed the `go` dir

andronat (approved):

PR #1269, opened 2026-06-19
feat(orb-connd): fix modem powercycle logic

pophilpo:
# Problem
 When ModemManager goes down, the modem supervisor detects the failure, powercycles the modem, and calls `restart_unit("ModemManager.service")`. This call only enqueues the systemd job — it returns immediately without waiting for MM to actually be up.

The supervisor then returns `Err`, and restarts after the minimum 10s backoff. On restart, the supervisor immediately calls `refresh_snapshot()` → `mmcli -L`. But `ModemManager` takes ~12s (or maybe longer) to fully initialize after being started. At the 10s mark MM is still starting, `mmcli -L` fails, and the supervisor concludes the modem is broken again — triggering a second unnecessary hardware powercycle that interrupts MM mid-startup.

This repeats with growing backoffs (10s, 20s, 30s…), causing repeated unnecessary powercycles and making ModemManager recovery slow.

# Fix
`powercycle_modem` now polls the` ActiveState` of `ModemManager.service` after calling `restart_unit`, blocking for up to 30s until the unit is fully active before returning.

### DIFF
--- a/orb-connd/Cargo.toml
+++ b/orb-connd/Cargo.toml
@@ -7,7 +7,7 @@ license.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 authors = [
-	"Philipp Poppov <pophilpo@users.noreply.github.com>",
+	"Philipp Popov <pophilpo@users.noreply.github.com>",
 	"Victor Ferreira Menge <vmenge@users.noreply.github.com>",
 ]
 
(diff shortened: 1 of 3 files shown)


sky-mart (approved):

vmenge (approved):
noice!!

PR #1272, opened 2026-06-22
feat(prelude): log sampling, feature flagging

vmenge:
puts prelude features behind a feature flag and adds a log sampler

Logs are sampled per call site and level. The call site is identified by the macro invocation's file, line, and column.

The logging macros intentionally accept format-style log messages only. They do not support structured `tracing` fields such as `target:` or `field = %value`.

```rust
use std::time::Duration;
use prelude::logging::sampled::{self, LogSampler};

let mut logger = LogSampler::new(Duration::from_secs(1));

sampled::info!(logger, "waiting for {}", "capture");
sampled::warn!(logger, "retrying operation");
sampled::error!(logger, "operation failed: {}", "timeout");
```


pophilpo (approved):
🎰

PR #1277, opened 2026-06-22
feat(connd): primary conn metric

vmenge:
adds primary iface metrics

only supports wlan0, wwan0 and eth0 for now

### DIFF
--- a/orb-connd/src/reporters/connd_report.rs
+++ b/orb-connd/src/reporters/connd_report.rs
@@ -1,22 +1,27 @@
-use crate::network_manager::{Connection, NetworkManager};
+use crate::network_manager::NetworkManager;
 use color_eyre::{eyre::Context, Result};
 use flume::Receiver;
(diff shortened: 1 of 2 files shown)


pophilpo (approved):

vmenge:
> * [P1] `orb-connd/src/reporters/connd_report.rs:58`: This now derives `egress_iface` and the new primary-connection gauges from `oes::Connection.primary`, but that flag is inferred upstream from active connection names (`cellular`, `wired`, `ethernet`, or exact Wi-Fi SSID). A valid primary cellular/Ethernet connection with a different NetworkManager id will now report `egress_iface: None` and emit `0` for every interface, whereas the old code used `nm.primary_connection()` directly and would still report `wwan0`/`eth0`. Please keep using the authoritative NM primary connection here, or make the active-connections report carry an authoritative primary/type, and add a regression test for non-default connection names.

no

PR #1283, opened 2026-06-23
chore: remove experiments from workspace

vmenge:
journey to avoid headaches

ibuprofen and paracetamol, you better watch out


AlexKaravaev (approved):

vmenge:
> ### Blocking
> * `Cargo.toml:2` - Removing the experiment crates from `workspace.members` leaves their manifests under the workspace tree but no longer buildable. They still inherit workspace fields/dependencies via `*.workspace = true`, and Cargo now errors when invoked directly, e.g. `cargo metadata --manifest-path experiments/artificer/Cargo.toml` reports:
> 
> ```
> error: current package believes it's in a workspace when it's not
> ```
> 
> This applies to the removed experiment manifests under `experiments/artificer`, `experiments/zenoh`, and `experiments/orb-blob/*`. Either keep them as workspace members, delete them, or make them valid standalone packages with explicit package/dependency values or their own workspace setup.

image

PR #1291, opened 2026-06-26
feat(update-agent): update metrics

vmenge:
## new
metrics for:
- success of update application
- error causes
- claim endpoint success rate
- component update application duration (per manifest ver)

## other
- adds update-agent to zenohd group
- removes `UpdateAgentResult` in favour of using `ExitCode` directly
- moved reboot logic up to guarantee metrics are flushed

## tests
- [ ] on diamond orb


update-agent/src/claim.rs
```diff
@@ -133,11 +139,14 @@ fn from_remote(
...
 
+    let conn_stability = conn_tracker.track_stability();
     let client = crate::client::normal()?;
+    // attempt starts here
+    // ignore if we dont have internet
     let resp = client
         .get(api_url.clone())
         .send()
-        .map_err(Error::SendCheckUpdateRequest)?;
+        .map_err(Error::SendCheckUpdateRequest)?; // early exit here
```
pophilpo: Should we also include that in the failed metrics or something ?

update-agent/src/claim.rs
```diff
@@ -155,9 +164,20 @@ fn from_remote(
...
+                );
+            }
+
             Err(Error::status_code(status, msg))
         }
     } else {
+        let _ =
+            metrics.count("orb.platform.update-agent.claim-request", 1, ["ok:true"]);
+
         let resp_txt = resp.text().map_err(Error::ResponseAsText)?;
```
pophilpo: Should we count success after we actually deserialize the response? WHAT IF
we get something broken in the body, that would still be a success

update-agent/src/claim.rs
vmenge: unfortunately a lot of the errors there are mixed between client and server in a way that's hard to differentiate, so i'd rather not include them for now

update-agent/src/claim.rs
vmenge: yee, that's a good point.  i wanted to keep it minimal but would be good to make sure contract isn't broken

pophilpo (approved):

update-agent/src/lib.rs
```diff
@@ -9,20 +9,143 @@ pub mod settings;
...
+};
 
 pub const BUILD_INFO: BuildInfo = make_build_info!();
 
+mod exit_codes {
+    pub const DOWNLOAD_FAILED: u8 = 150;
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
```
sfikastheo, line 31: any good reason to move it out its separate module that it was under?

update-agent/src/main.rs
```diff
@@ -49,36 +42,145 @@ use orb_update_agent_dbus::{
...
+use zenorb::Zenorb;
 
 const CFG_DEFAULT_PATH: &str = "/etc/orb_update_agent.conf";
 const ENV_VAR_PREFIX: &str = "ORB_UPDATE_AGENT_";
 const CFG_ENV_VAR: &str = const_format::concatcp!(ENV_VAR_PREFIX, "CONFIG");
 const SYSLOG_IDENTIFIER: &str = "worldcoin-update-agent";
 
-fn main() -> UpdateAgentResult {
+#[tokio::main]
+async fn main() -> ExitCode {
```
sfikastheo, line 68: it happened :sadblob:

sfikastheo (approved):
Thats dope!!

update-agent/src/lib.rs
vmenge: this is a brand new error :sweat_smile:

update-agent/src/main.rs
vmenge: :kek:

pophilpo (approved):

sfikastheo (approved):
last change also makes sense to me. re-approving, but you can touchup on the clanker's finding above

PR #1293, opened 2026-06-29
feat(ci): self-contained cargo vulnerability scan

Cr0wTom:
Moved the scan internally from calling externally from orb-internal.

Tested in branch and working without errors.


sfikastheo (approved):
Love it <3

PR #1294, opened 2026-06-29
fix(bidiff-util): leave manifest data untouched

sfikastheo:
https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html
The de-serialization & re-serialization should not be touching the manifest data


pophilpo (approved):
👍🏻 🚀 🍊

AlexKaravaev (approved):

PR #1301, opened 2026-07-03
feat: specify the platform in update agent requests

sky-mart:
https://tfh.slack.com/archives/C052M1G8SMA/p1783075906881779

### DIFF
--- a/update-agent/src/claim.rs
+++ b/update-agent/src/claim.rs
@@ -9,6 +9,7 @@ use std::{
 
 use eyre::{ensure, WrapErr as _};
 use orb_dogd::{DogstatsdClient, MetricEmitter};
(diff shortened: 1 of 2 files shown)


update-agent/src/claim.rs
```diff
@@ -296,6 +297,7 @@ fn ensure_sources_match_claim(
 pub fn get(
     settings: &Settings,
     version_map: &VersionMap,
+    platform: &str,
```
sfikastheo: I would propagate [orbosplatform](https://github.com/worldcoin/orb-software/blob/main/orb-info/src/orb_os_release.rs#L24) then `into()` would translate it into the request at the last step to str

sfikastheo (approved):

PR #1305, opened 2026-07-09
feat(worldcoin-attest): Add ExecStartPre for permissions on /usr/persistent/se

BulatAbuzarov:
Added ExecStartPre commands to set permissions and ownership for /usr/persistent/se.

### DIFF
--- a/attest/debian/orb-attest.worldcoin-attest.service
+++ b/attest/debian/orb-attest.worldcoin-attest.service
@@ -21,6 +21,9 @@ Environment=DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/worldcoin_bus_socket
 Environment=RUST_BACKTRACE=1
 SyslogIdentifier=worldcoin-attest
 Restart=on-failure
+# In case /usr/persistent/se was manually re-proved under root
+ExecStartPre=+/bin/chmod -R +X /usr/persistent/se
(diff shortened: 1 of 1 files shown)


pophilpo (approved):

pophilpo (approved):

BulatAbuzarov:
> Found concrete issues that should block merge:
> 
> * `attest/debian/orb-attest.worldcoin-attest.service:26`: `chown -R worldcoin:worldcoin /usr/persistent/se` runs as root and follows symlink referents by default. Since `/usr/persistent/se` is created/used by the `worldcoin` user, that user can place symlinks in the tree and have arbitrary targets chowned to `worldcoin` on the next service start/reboot. Use non-dereferencing traversal, validate the top-level path is not a symlink, or avoid recursive root `chown` over user-controlled persistent content.
> * `attest/debian/orb-attest.worldcoin-attest.service:17-26`: the unit starts the daemon as `orb-attest`, but the new pre-start step changes `/usr/persistent/se` to `worldcoin:worldcoin`, and `orb-attest` is not in the `worldcoin` group. For restrictive files from the “re-proved under root” case, this still does not grant the daemon access and may leave signing broken. The ownership/mode should match the account that actually reads the SE material, or grant explicit group access.

traversal of symlink is disabled by default on chown

PR #1308, opened 2026-07-09
refactor(connd): use crabwire::inject for DI on modem_supervisor

vmenge:
## changes
- refactors modem supervisor and other functions that share the same dependencies to use `crabwire::inejct` to inject depndencies to avoid having to drill them down every single function

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2475,6 +2475,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "crabwire"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "436cde1b854d98a37baace56bf17cd64a4b48821851ccdfe648908c3e1c48666"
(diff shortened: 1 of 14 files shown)


pophilpo (approved):

vmenge:
> * **Blocking – test fixtures share stale/global dependencies** ([orb-connd/tests/fixture.rs:221](/home/runner/work/orb-software/orb-software/orb-connd/tests/fixture.rs:221)): `try_register!` silently retains the first process-global registry. Subsequent fixtures—often running concurrently—therefore use the first fixture’s `Systemd` connection and modem/MCU mocks. `restart()` also creates a new D-Bus connection without replacing the registered `Systemd`, leaving tasks connected to the removed container. Tests can become order-dependent, race, or exercise the wrong mocks. The registry must be isolated/reset per fixture, or dependencies should remain explicit in tests.

literally not relevant for any of the tests where this is used right now

PR #1312, opened 2026-07-10
chore(connd): attempt modem manager restart twice; modem supervisor tests

vmenge:
## changes
- attempts to restart modem manager twice when power cycling modem
- make `Systemd::restart_service` report on success of restart
- use `faux` for struct based mocking instead of `mockall`
- add tests for modem supervisor


vmenge:
> * **[P1] Integration tests are silently disabled** — [orb-connd/tests/modem_supervisor.rs:1](/home/runner/work/orb-software/orb-software/orb-connd/tests/modem_supervisor.rs:1): `cargo test --all --all-targets` does not enable the new `testing` feature, so this crate-level `cfg` removes all new modem tests—and similarly gated existing integration tests. Enable the feature in CI/default testing or avoid gating the test crates.

dumb clanker hallucinating

pophilpo (approved):

PR #1326, opened 2026-07-22
fix(attest): recreate client on every flow 

vmenge:
recreate client on every flow to avoid dns resolution issues when switching connections

still need testing on an orb


pophilpo (approved):

vmenge:
> * **[P1] Preserve the existing HTTP timeout** — `attest/src/client.rs:30` reduces every attestation request timeout from 60s to 30s. This is unrelated to client recreation and can cause token refreshes to fail on slow Orb connections that previously succeeded. Please retain 60s unless this behavioral change is independently justified and tested.

dont tell me what to do

PR #1331, opened 2026-07-23
fix(attest): also recreate client on get_working_static_token

vmenge:
```rust
        let token = select! {
            Ok(token) = get_working_static_token(orb_id, &ping_url) => token,
            token = remote_api::get_token(orb_id, &auth_url, &metrics, &conn_tracker) => token,
        };
```

in this loop if we don't recreate the http client in `get_working_static_token`, then `remote_api::get_token` gets an unfair bias due to the other function getting stuck with a bad http client for all retries as opposed to `remote_api::get_token` which recreates on every challenge flow attempt


sfikastheo:
artisanal review >> clanker

sfikastheo (approved):

PR #1335, opened 2026-07-30
fix: Orb mini Configuration for github-runner and jenkins-runners and flashing rig

chrisgalanis:
.

### DIFF
--- a/nix/machines/jenkins-agent.nix
+++ b/nix/machines/jenkins-agent.nix
@@ -39,6 +39,7 @@ let
     pkgs.android-tools
     qdl-rs
   ];
+  systemPath = "/run/wrappers/bin:/run/current-system/sw/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin";
 in
 {
   options.worldcoin.jenkinsAgent = {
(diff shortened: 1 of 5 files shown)


nix/machines/jenkins-agent.nix
```diff
@@ -12,6 +12,34 @@
...
+    pkgs.git
+    pkgs.bash
+    pkgs.nettools
+    pkgs.coreutils
+    pkgs.uv
+    python
+    pkgs.android-tools
+    qdl-rs
+  ];
+  systemPath = "/run/wrappers/bin:/run/current-system/sw/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin";
```
sfikastheo: Do we really need native linux paths to be exposed here?

nix/machines/jenkins-agent.nix
```diff
@@ -100,6 +128,7 @@ in
 
     systemd.tmpfiles.rules = [
       "d ${cfg.workDir} 0755 ${agentUser} ${agentUser} - -"
+      "d /home/${agentUser} 0755 ${agentUser} ${agentUser} - -"
```
sfikastheo: Wont just doing this
```
users.users.${agentUser} = {
  isNormalUser = true;
  home = "/home/jenkins-agent";
  createHome = true;
  extraGroups = [ "wheel" ] ++ cfg.extraGroups;
};
```

IN the user creation solve this? Do you want tmp on home? 
What is going to be there?

sfikastheo (approved):

PR #1337, opened 2026-08-03
feat(orb-monitoring): orb monitoring auth service

vmenge:
skeleton for the service with most functionality implemented
not ready for production, but ready to merge

will wait for backend server to be up for final adjustments


orb-monitoring-auth/src/server/serve_token.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        ctx.oneshot(async move |_| -> Result<()> {
+            let token = token
+                .read()
+                .map_err(|e| eyre!("{e:?}"))?
+                .deref()
+                .clone()
+                .filter(|token| !token.is_expired_at(clock.now()));
+            let payload = serde_json::to_vec(&token)?;
+            stream.write_all(&payload).await?;
```
pophilpo, line 52: `oneshots` insert every child’s `AbortHandle` into the context’s children map, it never removes that entry when the task completes. Every  request retains a completed task entry  until shutdown or listener restart, causing potentially unbounded memory growth or at least slower `Drop` for that `ctx` 🤷🏻‍♂️ 
Not sure how practical is this and if this will ever grow to something meaningful in the field, but still maybe a thing to look at

orb-monitoring-auth/src/client/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    unsafe {
+        let uid = getuid();
+        if uid != dd_agent_uid {
+            bail!(
+                "user with id {uid} is not allowed to call orb-monitoring-auth-client"
+            )
+        }
+    }
+
```
pophilpo, line 74: We can avoid `unsafe` block here with `let uid = nix::unistd::getuid().as_raw();`
But this requires a "user"  feature in the nix crate.

Or use `rustix::process::getuid().as_raw();`

orb-monitoring-auth/src/client/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    unsafe {
+        let uid = getuid();
+        if uid != dd_agent_uid {
+            bail!(
+                "user with id {uid} is not allowed to call orb-monitoring-auth-client"
+            )
+        }
+    }
```
pophilpo, line 73: We can avoid `unsafe` block here with `let uid = nix::unistd::getuid().as_raw();`
But this requires a "user" feature in the nix crate.

Or use `rustix::process::getuid().as_raw();`

orb-monitoring-auth/src/server/serve_token.rs
vmenge: eh, we should have this being hit once or twice per boot
i mean even if it was hit 100 times in a power cycle it wouldnt even be 100kB of RAM, so i think we're fine :sweat_smile:

orb-monitoring-auth/src/server/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+    fn name(&self) -> &'static str {
+        "orb-monitoring-auth-server"
+    }
+
+    fn about(&self) -> &'static str {
+        "fetches monitoring token from our backend, stores it in SecureStorage and serves it to datadog-agent"
+    }
+
+    fn main(&self, _args: &ArgMatches) -> ExitCode {
+        let result = tokio::runtime::Builder::new_multi_thread()
```
alekseifedotov, line 45: Why use tokio here? This service has 1 client at most.

vmenge: 1 client, but it also concurrently refreshes token in background if needed.

but overall i just like tokio and i like the ecosystem around it which makes my life easier. i get retry logic for free with `speare` which uses `tokio`, and end up having to write less logic, test less logic and have an easier time maintaining this

pophilpo (approved):

PR #1344, opened 2026-08-12
feat(orbd): run once 15m after boot

pophilpo:
Change to running once per boot, 5m after boot

### DIFF
--- a/orbd/debian/worldcoin-orbd.timer
+++ b/orbd/debian/worldcoin-orbd.timer
@@ -1,9 +1,8 @@
 [Unit]
-Description=Run Worldcoin Orb Daemon every 15 minutes
+Description=Run Worldcoin Orb Daemon 15 minutes after boot
 
 [Timer]
-OnBootSec=3min
-OnUnitInactiveSec=15min
+OnBootSec=15min
 
 [Install]
 WantedBy=timers.target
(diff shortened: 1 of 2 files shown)


orbd/debian/worldcoin-orbd.timer
```diff
@@ -1,9 +1,8 @@
 [Unit]
-Description=Run Worldcoin Orb Daemon every 15 minutes
+Description=Run Worldcoin Orb Daemon 15 minutes after boot
 
 [Timer]
-OnBootSec=3min
-OnUnitInactiveSec=15min
+OnBootSec=15min
```
AlexKaravaev: let's do 5 minutes

AlexKaravaev (changes requested):

orbd/debian/worldcoin-orbd.timer
pophilpo: Sure

AlexKaravaev:
@pophilpo is the codex comment valid?

pophilpo:
> * **Blocking – existing installations will still start `orbd` at boot:** Removing `WantedBy=multi-user.target` from [worldcoin-orbd.service](orbd/debian/worldcoin-orbd.service#L18) does not remove the existing `multi-user.target.wants/worldcoin-orbd.service` symlink created by previous package versions. Upgraded Orbs can therefore run it immediately at boot and again via the five-minute timer. Add a package migration that explicitly disables the formerly enabled service while retaining the timer.

No existing installations are present

AlexKaravaev (approved):

PR #1346, opened 2026-08-13
fix(backend-status): return version from /etc/os-release

sfikastheo:
Do not use `version.json`


orb-backend-status/src/main.rs
```diff
@@ -58,7 +57,7 @@ async fn main() -> Result<()> {
         .dbus(zbus::Connection::session().await?)
         .zsession(&zsession)
         .endpoint(endpoint)
-        .orb_os_version(orb_os_version()?)
+        .orb_os_version(OrbOsRelease::read().await?.orb_os_version)
         .orb_id(orb_id)
```
pophilpo, line 61: Should you use here the `platform_version` helper?

I think the backend currently displays something like this: 
`7.0.0-diamond-prod` and the `.orb_os_version` will report only `7.0.0`

There's a method you created actually :D 

```

pub fn platform_version(&self) -> String {
        format!(
            "{}-{}-{}",
            self.orb_os_version, self.orb_os_platform_type, self.release_type
        )
    }
```

AlexKaravaev: :this: ⏫  👍 ➕ 💯

sfikastheo: Thanks missed it. Patching

AlexKaravaev (approved):

PR #1354, opened 2026-08-20
chore(deny): fix deny

pophilpo:
Fix deny

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
  "bytes",
- "crypto-common",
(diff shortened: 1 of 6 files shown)


pophilpo:
> * **[P1] [Cargo.toml:93](/home/runner/work/orb-software/orb-software/Cargo.toml:93)** — The AWS features enable `rustls`’s `aws_lc_rs` provider while the workspace already enables `ring`. The lockfile confirms both providers are compiled. Without explicitly installing one provider, `rustls` cannot choose and reqwest/AWS client construction can panic at runtime. Select a single workspace-wide provider or install one before creating TLS clients.

Reqwest 0.12.24 explicitly falls back to `Ring`

AWS Smithy explicitly selects `CryptoMode::AwsLc`

Although still valid to fix in the future, this can cause issues

alekseifedotov:
There is already PRs which do the same

https://github.com/worldcoin/orb-software/pull/1355
https://github.com/worldcoin/orb-software/pull/1349

PR #1367, opened 2026-08-26
fix: start devcontainer without claude

alekseifedotov:
Fixup for PR: https://github.com/worldcoin/orb-software/pull/1347 

If the user never run claude on their host, the container creating will fail without this PR

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -22,7 +22,7 @@
 			]
 		}
 	},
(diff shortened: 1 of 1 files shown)


sfikastheo (approved):
Yep, this was a comment in the previous PR, thanks for patching it

alekseifedotov:
> Yep, this was a comment in the previous PR, thanks for patching it

I was meaning to fix it in the previous PR but it had audo-merge enabled, so it was merged before I had a chance. Thanks for approving this PR.

PR #1368, opened 2026-08-27
feat: add a command to check if the crate is supported

alekseifedotov:
Not all crates build on all architectures(targets), some crates specify `[package.metadata.orb] unsupported_target` in Cargo.toml . To check if a crate is supported on a specific target, add a command `cargo x crate-supported`

### DIFF
--- a/.github/workflows/rust-ci.yaml
+++ b/.github/workflows/rust-ci.yaml
@@ -152,11 +152,18 @@ jobs:
 
       - name: Configure cargo to exclude platform-specific crates
         run: |
-          tmp=($(ci/rust_ci_helper.py excludes))
+          crates="$(cargo metadata --no-deps --format-version=1 | jq -r '.packages[].name')"
           EXCLUDES=""
-          for e in ${tmp[@]}; do
-            EXCLUDES+="--exclude ${e} "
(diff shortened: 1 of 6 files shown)


.github/workflows/rust-ci.yaml
```diff
@@ -147,11 +147,11 @@ jobs:
...
       - name: Configure cargo to exclude platform-specific crates
         run: |
-          tmp=($(ci/rust_ci_helper.py excludes))
+          crates="$(cargo metadata --no-deps --format-version=1 | jq -r '.packages[].name')"
           EXCLUDES=""
-          for e in ${tmp[@]}; do
-            EXCLUDES+="--exclude ${e} "
-          done
+          while IFS= read -r c; do
+            cargo x crate-supported "$c" >/dev/null || EXCLUDES+="--exclude ${c} "
```
sfikastheo: Should you not exclude packages only when the exit status is `1`. 
You will hide any random failure of `cargo x crate-supported` with `||`

alekseifedotov: Fixed.

sfikastheo (approved):

PR #1372, opened 2026-08-27
refactor: simplify OrbStatusApiV2 structure

alekseifedotov:
The outer `Option` is not necessary, if vector of events is empty, that is the same as `oes == None`.

### DIFF
--- a/oes/README.md
+++ b/oes/README.md
@@ -63,7 +63,7 @@ the same `OrbStatusApiV2` schema, but only the `oes` field is populated:
 ```rust
 struct OrbStatusApiV2 {
     oes_cached: bool, // true only whenever cached events are sent (every 30s)
-    oes: Option<Vec<Event>>,
+    oes: Vec<Event>,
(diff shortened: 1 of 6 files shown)


vmenge:
i'm not a fan tbh, i prefer the explicitness of `Option` there on the type as opposed to the magic of the macro, which when constructing the type will not make it immediately evident that might be missing from the payload (which matches the backend type as well) :thinking: 

but dont feel too strongly about it. did you test with an orb / were events still received without issue in the backend?

pophilpo:
I agree with @vmenge  here (terminal Rust brainrot situation 🦀).

An empty `vec![]` makes it less explicit that the field will be **omitted** from the payload entirely, whereas `None` communicates that directly. 
The backend also uses `Option<Vec<...>>` for this field, representing the same “field may be absent” concept. Although I should note that it currently accepts a missing field, so this does not affect compatibility, but I think the explicit type better documents the payload contract.

PR #1381, opened 2026-09-02
feat(connd): ble advertisement

vmenge:
## new 
- ble advertiser task, supporting concurrent advertisements

## changes
- `OrbCapabilties` to be a struct instead of enum and support `bluetooth` field

## notes
- connd supports an unlimited number of bt advertisements
- tested with up to 5 advertisements concurrently without any issues performing the signup E2E flow with the demo android apk for world id
- in orbs where bluetooth is not available or supported the ble advertisement task will simply not run

## tests
- tested manually on an orb with bt support
- integration tests on the pr
- tested on orbs that dont support bt with [a hil run](https://github.com/worldcoin/orb-os/pull/2535)

### DIFF
--- a/orb-connd/src/ble/mod.rs
+++ b/orb-connd/src/ble/mod.rs
@@ -1,0 +1,121 @@
+use bluer::adv::{Advertisement, Type};
+use color_eyre::{eyre::eyre, Result};
+use serde::{Deserialize, Serialize};
+use speare::mini;
+use std::collections::BTreeMap;
+use tracing::{error, info, warn};
+use uuid::Uuid;
+use zenorb::Zenorb;
+
+pub struct Args {
+    pub zenoh: Zenorb,
+}
+
+#[derive(Serialize, Deserialize)]
+struct Advert {
+    service_id: Uuid,
+    payload: Option<Vec<u8>>,
+}
+
+pub async fn advertiser(ctx: mini::Ctx<Args>) -> Result<()> {
+    info!("starting ble advertiser");
+
+    let session = bluer::Session::new()
+        .await
+        .inspect_err(|e| warn!("failed to create bluer session: {e}"))?;
+
+    let adapter = match session.default_adapter().await {
+        Err(bluer::Error {
+            kind: bluer::ErrorKind::NotFound,
+            ..
+        }) => {
+            warn!("no bluetooth adapter found. ble advertiser task will quit early.");
+            return Ok(());
+        }
+
+        Err(e) => {
+            error!("failed to acquire default ble adapter: {e:?}");
+            return Err(e.into());
+        }
+
+        Ok(a) => a,
+    };
+
+    let active = adapter.active_advertising_instances().await?;
+    let supported = adapter.supported_advertising_instances().await?;
+
+    info!("ble advertising instances. active: {active}, supported: {supported}");
+
+    if !adapter
+        .is_powered()
+        .await
+        .inspect_err(|e| warn!("failed to check ble adapter power status: {e}"))?
+    {
+        adapter
+            .set_powered(true)
+            .await
+            .inspect_err(|e| warn!("failed to power ble adapter: {e}"))?;
+    }
+
+    let subscriber = ctx
+        .zenoh
+        .declare_subscriber("ble_beacon")
+        .await
+        .map_err(|e| eyre!("{e}"))?;
+
+    let mut service_data = BTreeMap::new();
+    let mut _advertisement_handle = None;
+
+    loop {
+        let sample = subscriber
+            .recv_async()
+            .await
+            .map_err(|e| eyre!("{e}"))
+            .inspect_err(|e| {
+                warn!("ble advertiser failed receive zenoh sample: {e}")
+            })?;
+
+        let payload = sample.payload().to_bytes();
+        let advert = match serde_json::from_slice::<Advert>(&payload) {
+            Err(e) => {
+                warn!("ble advertiser received malformed json: {e}");
+                continue;
+            }
+
+            Ok(p) => p,
+        };
+
+        match advert.payload {
+            None => {
+                service_data.remove(&advert.service_id);
+            }
+
+            Some(payload) => {
+                service_data.insert(advert.service_id, payload);
+            }
+        }
+
+        match (service_data.is_empty(), &_advertisement_handle) {
+            (true, None) => (),
+
+            (true, Some(_)) => {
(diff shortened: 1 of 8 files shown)


vmenge:
image

`libdbus-sys` [is MIT](https://github.com/diwic/dbus-rs/blob/master/LICENSE-MIT)
`bluer` uses GPL only for its documentation, [the crate itself is BSD.](https://github.com/bluez/bluer?tab=License-1-ov-file)

orb-connd/src/ble/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+                }
+
+                // Dropping the handle stops advertising asynchronously. The next advertisement
+                // may briefly overlap with the old one; our module supports up to 8 at once.
+                drop(advertisement_handle.take());
+
+                let advertisement = Advertisement {
+                    advertisement_type: Type::Broadcast,
+                    // Nice to have only: BlueZ may truncate this if ServiceData needs the room.
+                    local_name: Some("Orb".to_owned()),
```
Cr0wTom: This means that the advertisement is scannable and that nearby devices can send SCAN_REQ packets, which the Orb will have to process. 

We can remove it to limit this attack vector, and the app/phone can discover the BLE advertisements using service identifiers.

Cr0wTom (changes requested):
Please check the comment and let's discuss it if needed. there are a couple of places that can result in availability issues in extreme scenarios, but let's leave them for now.

Cr0wTom:
@codex

AlexKaravaev:
@codex review

Cr0wTom (approved):
Please check the reliability findings from Codex @vmenge before merging

orb-connd/src/ble/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+        .await
+        .inspect_err(|e| warn!("failed to create bluer session: {e}"))?;
+
+    let adapter = match session.default_adapter().await {
+        Err(bluer::Error {
+            kind: bluer::ErrorKind::NotFound,
+            ..
+        }) => {
+            warn!("no bluetooth adapter found. ble advertiser task will quit early.");
+            return Ok(());
```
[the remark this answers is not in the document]
vmenge: no

orb-connd/src/ble/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+                drop(advertisement_handle.take());
+
+                let advertisement = Advertisement {
+                    advertisement_type: Type::Broadcast,
+                    service_data: BTreeMap::from([(service_id, payload.clone())]),
+                    timeout: Some(Duration::ZERO),
+                    ..Default::default()
+                };
+
+                advertisement_handle = Some(adapter.advertise(advertisement).await?);
```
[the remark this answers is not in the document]
vmenge: no

orb-connd/tests/fixture/mod.rs
```diff
@@ -149,6 +164,29 @@ impl Fixture {
...
+                Launcher::daemon()
+                    .bus_type(BusType::Session)
+                    .launch()
+                    .expect("failed to launch Bluetooth test bus")
+            })
+            .await
+            .unwrap();
+
+            unsafe {
+                env::set_var("DBUS_SYSTEM_BUS_ADDRESS", bus.address());
```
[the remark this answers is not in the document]
vmenge: no

vmenge:
> Please check the reliability findings from Codex @vmenge before merging

all duds, i always check them locally as well beforehand and have the same results

orb-connd/debian/worldcoin-connd.service
```diff
@@ -2,6 +2,7 @@
 Description=Worldcoin Connection Daemon
 Requires=NetworkManager.service
 After=NetworkManager.service
+After=bluetooth.service
```
AlexKaravaev, line 5: I think this will break on Pearl? We don't enable bluetooth there?

AlexKaravaev: i just don't know what `after` will  do in this case, i know it shouldn't fail if bluetooth.service fails but have no idea what happens if it's not present at all

AlexKaravaev (approved):

orb-connd/debian/worldcoin-connd.service
vmenge: it just ignores it if not present :catyes:

in [the docs](https://man7.org/linux/man-pages/man5/systemd.unit.5.html) it says:
> those settings are independent of and orthogonal to the requirement dependencies as configured by Requires=, Wants=, Requisite=, or BindsTo=

i did run it through the HIL just to be sure, no issues :D

orb-connd/src/ble/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+                    continue;
+                };
+
+                let advertisement_unchanged = last_service_id == Some(service_id)
+                    && last_payload.as_ref() == Some(payload);
+
+                if advertisement_unchanged {
+                    continue;
+                }
+
```
pophilpo, line 138: Should we invalidate these if `Bluez` restarts for some reason ? It looks like we will not try to advertise in that scenario, basically loosing this advert.

pophilpo (approved):
Looks good to me. Left one comment, not sure if it's valid here though. Please enlighten me!

orb-connd/src/ble/mod.rs
vmenge: > Should we invalidate these if Bluez restarts for some reason
wdym by invalidate?

but i let the advertisements be lost on restart because once we're publishing it could be the error is happening because of contents of the advertisement's payload, so just to be sure i let them be cleared

should "never" happen theoretically, but even if it does, current producer (`orb-core`) and future planned ones would shortly publish the payload again anyway, so would just result in a small blip (if error is due to some other cause)

vmenge:
> * **[P2] Preserve advertisements across registration failures** — [orb-connd/src/ble/mod.rs:150](orb-connd/src/ble/mod.rs#L150). Any `adapter.advertise()` error exits the task, discarding every service’s payload. The supervisor restarts with an empty map, and the Zenoh subscriber does not replay earlier publications, so a transient Bluetooth failure permanently stops existing beacons until producers republish. Retain the desired advertisements while retrying registration, and add a recovery test that succeeds without republishing.
> 
> Tests were not run in the read-only environment.

stop

dumb bot please read previous replies before bothering me

PR #1384, opened 2026-09-03
fix: orb name on Android

alekseifedotov:
Add implementation for getting the orb name on Android. The name is derived from orb-id, there is 1:1 match between them.

for example orbid `666666` always has name `appear-fabric-abandon` and you can find orb id from the name without looking up any database. 

NOTE: orb-ids for mini are 3 words, not 2 like for diamond and pearl, that will help us distinguish them at a glance.


sfikastheo (approved):
This is dope i think we should merge it

PR #1385, opened 2026-09-04
feat: refactor tests

alekseifedotov:
A bunch of small fixes:
- add function `test_orb_id()`  returing  a pre-defined orb-id to be used only in tests.
- use OrbId instead of &str for orb-id
- fix a race in a zenorb test


zenorb/tests/it_werks.rs
```diff
@@ -116,6 +116,16 @@ async fn it_werks() {
         .await
         .unwrap();
 
+    // put().await only confirms the message was sent, not that the remote
```
sfikastheo, line 119: I have a suspicion this was on purpose? 
@vmenge should review it. 
Everything else looks very good to me.

alekseifedotov: Yes, that was on purpose. I think there is a race in the test, and this change tries to fix it.

vmenge: you mean with the expected messages arriving out of order? 

i've never seen this test be flaky in our pipeline but it is a real possibility 🤔 

do you have a link to a run where this happened?

though if we add this check we should add it after the first publish as well right? otherwise we will still have a race condition between `bytestopic` and `texttopic`

vmenge (changes requested):

zenorb/tests/it_werks.rs
vmenge: while a race condition is possible on the tests, i don't remember ever seeing a failure bc of it tbh, not a blocker but would appreciate some links to runs that failed bc of it just to take a look 🙏

vmenge (changes requested):
on the changes of `it_werks.rs` we should wait for `received_msgs` to be incremented after every send, not only before the querying, otherwise the race condition remains

pophilpo (approved):

vmenge (approved):

PR #1387, opened 2026-09-04
fix: clippy warning use of `default` to create a unit struct

alekseifedotov:
Fix:
image


vmenge (approved):

vmenge:
btw @alekseifedotov you might be a running a more strict clippy or different version than the one in the repo :thinking: 

could u share what you're running? maybe we can add it to our clippy setup in the pipeline too

alekseifedotov:
>btw @alekseifedotov you might be a running a more strict clippy or different version than the one in the repo 🤔

>could u share what you're running? maybe we can add it to our clippy setup in the pipeline too

I'm using a devcontainer, so whatever is installed there.

I'm surprised that it gives different results between my devcontainer and CI, isn't nix supposed to be reproducible?

PR #1392, opened 2026-09-04
fix(jobs-agent): restrict zoci command tokens to a safe charset

Cr0wTom:
The zoci fallback interpolates the job document's first token into a Zenoh key expression, so `*`/`**` fanned one relay job out to every internal `job/*` queryable (connd wifi_*, supervisor gondor).

Fix: `get_zoci_command` accepts only 1–64 byte tokens of `[a-z0-9_]` (all six real endpoints conform; unknown commands still get `FailedUnsupported`), with a bounded warn on rejection.

Tests: accept/reject matrix, green natively (`cargo test -p orb-jobs-agent --lib`).

Mitigates: wildcard portion of VULN-6316 (prefix-dispatch portion stays open on the ticket); external disclosure report (Sep 2026), zoci fan-out finding.


vmenge (approved):
lgtm

would be better without clanker comments 👀

PR #1393, opened 2026-09-04
fix(jobs-agent): validate relay message sender before dispatch

Cr0wTom:
The jobs agent dispatched relay `JobNotify`/`JobExecution`/`JobCancel` without checking `msg.from` — any relay entity that can address the Orb in the jobs namespace could run privileged jobs or cancel live ones.

Fix: new `is_authorized_sender` (Service + configured `target_service_id` + configured `relay_namespace`) evaluated before any payload decode; anything else is dropped with a bounded, escaped rejection log (payload never logged). Fleet-safety basis: the relay routes by identity triple and a client sends with the identity it connected as, so the job-server provably owns the configured triple — otherwise the agent's own `JobRequestNext` would never reach it today. Sender-label authenticity itself is relay-server work (SEC-2708).

Tests: authorization matrix + payload classification units; adversarial integration tests (second relay client as `unexpected-service`, and `fleet-cmdr` with the wrong entity type) assert the rejection log fires, the rejected job never runs, and the agent keeps serving the legitimate server. Full suite green natively.

Mitigates: VULN-6325 (fixed); external disclosure report (Sep 2026), sender-validation finding. Complements SEC-2708.

### DIFF
--- a/orb-jobs-agent/src/job_system/client.rs
+++ b/orb-jobs-agent/src/job_system/client.rs
@@ -8,11 +8,27 @@ use orb_relay_messages::{
     jobs::v1::{
         JobCancel, JobExecution, JobExecutionUpdate, JobNotify, JobRequestNext,
     },
-    prost::{Message, Name},
+    prost::{DecodeError, Message, Name},
     prost_types::Any,
-    relay::entity::EntityType,
+    relay::{entity::EntityType, Entity},
 };
+use std::sync::atomic::{AtomicU64, Ordering};
 use tracing::{error, info, warn};
+
+/// Sender labels are attacker controlled and unbounded; truncate before logging.
+const MAX_LOGGED_LABEL_CHARS: usize = 128;
+
+/// Log the first few unexpected senders, then only every
+/// `UNEXPECTED_SENDER_LOG_EVERY`-th one, so warn volume never scales 1:1 with
+/// unsolicited traffic. NOTE: the counter is global, not per-sender -- a noisy
+/// unexpected sender can delay first-log evidence of a second distinct one;
+/// accepted for the shadow window (a per-sender map would grow unbounded on
+/// attacker-chosen keys).
+const UNEXPECTED_SENDER_LOG_BURST: u64 = 10;
+const UNEXPECTED_SENDER_LOG_EVERY: u64 = 100;
+
+/// Process-local count of inbound messages from unexpected senders.
+static UNEXPECTED_SENDERS: AtomicU64 = AtomicU64::new(0);
 
 #[derive(Debug, Clone)]
 pub struct JobClient {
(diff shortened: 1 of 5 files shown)


orb-jobs-agent/src/job_system/client.rs
```diff
@@ -8,12 +8,28 @@ use orb_relay_messages::{
...
+/// `UNEXPECTED_SENDER_LOG_EVERY`-th one, so warn volume never scales 1:1 with
+/// unsolicited traffic. NOTE: the counter is global, not per-sender -- a noisy
+/// unexpected sender can delay first-log evidence of a second distinct one;
+/// accepted for the shadow window (a per-sender map would grow unbounded on
+/// attacker-chosen keys).
+const UNEXPECTED_SENDER_LOG_BURST: u64 = 10;
+const UNEXPECTED_SENDER_LOG_EVERY: u64 = 100;
+
+/// Process-local count of inbound messages from unexpected senders.
+static UNEXPECTED_SENDERS: AtomicU64 = AtomicU64::new(0);
```
vmenge, line 31: please move this into `JobClient` or elsewhere (with some encapsulation), but please do not leave it as global mutable state 🙏 

probably would make sense to have a struct that abstracts the logic of counting senders and also throttles the logging, and then embedding that struct in `JobClient`

orb-jobs-agent/src/job_system/client.rs
```diff
@@ -218,6 +230,105 @@ impl JobClient {
...
+) -> bool {
+    // prost exposes `entity_type` as a raw i32; compare the same way the relay
+    // client itself does.
+    EntityType::try_from(from.entity_type) == Ok(EntityType::Service)
+        && from.id == target_service_id
+        && from.namespace == relay_namespace
+}
+
+/// Decodes an inbound relay message and classifies it by payload type.
+fn classify_inbound(payload: &[u8]) -> InboundDecision {
```
vmenge, line 265: mostly a nitpick:
would be more idiomatic to make this a `From<&[u8]>` impl for `InboudDecision`

orb-jobs-agent/src/job_system/client.rs
```diff
@@ -218,6 +230,105 @@ impl JobClient {
...
+            },
+        }
+    } else {
+        InboundDecision::UnknownType(any.type_url)
+    }
+}
+
+/// Logs a rejected message -- never its payload -- and only for the first few
+/// occurrences plus every `UNEXPECTED_SENDER_LOG_EVERY`-th one thereafter.
+fn log_unexpected_sender(from: &Entity) {
```
vmenge, line 307: function is  called `log_*` but also executes a side effect mutating global state 👀 
would ask to please rename fn and also not use global mutable state 🙏 
(moving state into `JobClient` struct would be fine, but other solutions ok as well just not global mutable state plz 🙏)

orb-jobs-agent/src/job_system/client.rs
```diff
@@ -42,76 +58,72 @@ impl JobClient {
...
+            // their payload is decoded.
+            if !is_authorized_sender(
+                &msg.from,
+                &self.target_service_id,
+                &self.relay_namespace,
+            ) {
+                log_unexpected_sender(&msg.from);
+                continue;
+            }
+
```
vmenge, line 80: i'm not sure if the lack of a `Self::request_next_job` call here might be problematic. could you add an integration test verifying we can still receive new job requests even after receiving an unauthorized sender? 🙏

vmenge (changes requested):
valve plz fix
image

orb-jobs-agent/src/job_system/client.rs
Cr0wTom: [e77a9d1](https://github.com/worldcoin/orb-software/pull/1393/commits/e77a9d19800d1322275d20f91a7c20e00ace70fb)

orb-jobs-agent/src/job_system/client.rs
Cr0wTom: [e77a9d1](https://github.com/worldcoin/orb-software/pull/1393/commits/e77a9d19800d1322275d20f91a7c20e00ace70fb)

orb-jobs-agent/src/job_system/client.rs
Cr0wTom: [e77a9d1](https://github.com/worldcoin/orb-software/pull/1393/commits/e77a9d19800d1322275d20f91a7c20e00ace70fb)

orb-jobs-agent/src/job_system/client.rs
Cr0wTom: [e77a9d1](https://github.com/worldcoin/orb-software/pull/1393/commits/e77a9d19800d1322275d20f91a7c20e00ace70fb)

PR #1397, opened 2026-09-04
fix(orb-connd): nm state cleanup

pophilpo:
orb-connd runs as orb-connd (sorry), but `NetworkManager` creates its persistent directories as `root:root 0700.` This caused startup cleanup to fail with `Permissiondenied`, allowing lease files to accumulate.

This change:

  - Repairs directory permissions before each connd start, preserving individual file ownership and modes. ( we kinda do the same in attest)
  - Measures allocated disk space rather than file contents.
  - Removes leases and seen-bssids before considering saved-profile eviction.
  - Tolerates missing files and does not fail on that

Filesystem operations are extracted into `nm_state.rs` but feel free to argue that it shouldn't.


pophilpo:
> * **High — symlink traversal can escape the state directory or loop forever:** [`orb-connd/src/service/nm_state.rs:35`](https://github.com/worldcoin/orb-software/blob/0c69c89d7fca07c3a0fa9d900957af6aec6c3f53/orb-connd/src/service/nm_state.rs#L35) uses `DirEntry::metadata()`, which follows symlinks. A directory symlink is consequently counted and pushed for recursive traversal at line 52. A link to a parent can make startup cleanup loop indefinitely; a link outside the NM directory causes unrelated files to be counted. Use `symlink_metadata()` or `file_type()` and never enqueue symlinks. The new “without following symlinks” test should also assert traversal behavior directly, as its current size-only assertion does not reliably catch this.

image

vmenge (approved):
very noice

PR #1402, opened 2026-09-04
fix: dont try to run flashing rig service if it is not there

alekseifedotov:
All orb-hils are trying to start flashing-rig service, even if it is not installed on the device, that creates noise in the log.

image

Make the `flashing-rig` service conditional and only enable it where it is needed.


AlexKaravaev (approved):
👍

PR #1410, opened 2026-09-14
feat(ipcp): add shared HPKE library (ORBA-682)

melmols:
[ORBA-682](https://linear.app/worldcoin/issue/ORBA-682)

## Changes

- Add `orb-ipcp-hpke` in `orb-software` for shared iPCP encryption, decryption and payload hashing.
- Use the agreed HPKE Base / X25519 / HKDF-SHA256 / AES-256-GCM profile and payload encoding.
- Add 12 inline Rust unit tests calling production functions, including known vectors, malformed payloads, tampering and entropy failure.
- Oxide and Orb-core integration are out of this PR.

## Validation

- All 12 Nextest tests and Clippy pass in the existing Linux container with the declared registry dependencies.
- The local cleanup-patched dependencies also pass all 12 tests, Clippy and optimized tests.
- Formatting and whitespace checks pass.


ipcp-hpke/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "orb-ipcp-hpke"
+version = "0.1.0"
+description = "Shared iPCP HPKE encryption and decryption"
+publish = false
+edition = "2024"
+rust-version = "1.93"
+license = "MIT OR (Apache-2.0 WITH LLVM-exception)"
```
pophilpo: We use top-level `Cargo.toml` for centralized configuration. You should use the same thing other crates use:

`rust-version.workspace = true
`

Also, that helps avoiding missmatch in the version, because `orb-software` is on `1.97.1`

Cargo.toml
```diff
@@ -19,6 +19,7 @@ members = [
 	"header-parsing",
 	"hil",
 	"io-utils",
+	"ipcp-hpke",
 	"jwk-util",
```
pophilpo, line 23: This adds a workspace member but I don't see  the Cargo.lock change commited. Please include it: without it, `--locked` builds fail and Cargo must resolve the new dependency graph locally.

pophilpo (changes requested):

Cargo.toml
pophilpo: NVM, saw that it's in progress to add

pophilpo (approved):

PR #1413, opened 2026-09-15
feat(telemetry): add Android support

pophilpo:
# Summary

This PR adds Android support to `orb-telemetry`

Android support is selected by the compilation target:

``` rust
    // Example usage
    // Remove the build target, if the service if fully-for android though
    // just 'with_logcat' is enough
    #[cfg(target_os = "android")]
    let telemetry = telemetry.with_logcat(c"orb-backend-status");
```

# Limitations
  - Android’s 4068-byte payload includes the priority, tag, message, and terminators. Messages are split into UTF-8-
    safe chunks of at most 4000 bytes. Tags exceeding 65 bytes are truncated with a warning in this impl instead of failing

- Logcat does not support ANSI colors and can have jumbo-mumbo text, so this implementation disables the colors
- Delivery is best-effort: Android’s logging API does not expose underlying logd transport failures

# Testing
Tested the Android path manually. Didn't test the linux path with these changes, but I think this should be fine, since all the changes are gated by target os.


vmenge (approved):

telemetry/src/logcat.rs
```diff
@@ -0,0 +1,326 @@
...
+        let chunk = CString::new(chunk).expect("NUL bytes were escaped above");
+
+        emit(&chunk)?;
+        rest = tail;
+    }
+
+    Ok(())
+}
+
+#[cfg(target_os = "android")]
```
alekseifedotov, line 72: The whole file could be under `#[cfg(target_os = "android")]` instead of having every function in the file compiled conditionally

pophilpo: The main reason for this was to enable host-side unit tests without requiring Android build target. I could improve this actually

pophilpo: I will follow up on this PR and fix this

PR #1414, opened 2026-09-15
fix: append 'arkenstone' sufix to orb id

alekseifedotov:
On Linux based orbs and Android based devices, the orb-id comes from different places and there is a potential for a collision. To avoid the collision *append* 'arkenstone' to the orb id on Android based orbs.

The backend likes to truncate orb id to 8 character when displaying it, so it is better to append the suffix to keep the more of the id visible.

### orb id example: 876120561arkenstone

Do you think it looks good? Any comments?


sfikastheo:
TBH, i do not think that this is necessary.

> The backend likes to truncate orb id to 8 character when displaying it, so it is better to append the suffix to keep the more of the id visible.

Why not just change this?

PR #1415, opened 2026-09-16
refactor(ipcp): share structured image HPKE payload

melmols:
[ORBA-665](https://linear.app/worldcoin/issue/ORBA-665) · [ORBA-684](https://linear.app/worldcoin/issue/ORBA-684)

## Changes

- Reuse the relay protobuf `IpcpHpkePayload` as `EncryptedPayload`; remove the duplicate payload type and hash helper.
- Keep one `PairingKey` with `new()`, `encrypt()` and `decrypt()`. Store the private key internally and expose HPKE's typed public key as `pk`; encryption takes the recipient's typed public key.
- Require `Zeroizing<Vec<u8>>` as encryption input and return it from decryption. Callers move the plaintext guard into the library without unwrapping it.
- Use HPKE’s `single_shot_seal()` / `single_shot_open()` to handle the combined ciphertext and authentication tag. Enable the existing `alloc` feature and remove manual tag parsing, appending and scratch-buffer management.
- Keep encryption/decryption synchronous with empty HPKE info/AAD internally. HPKE handles randomness; scheduling stays with callers. The HPKE suite and protobuf encoding are unchanged.
- Remove role state, extra constructors/accessors, raw private-key APIs and the direct Tokio dependency.
- Merge before [Orb-core #2174](https://github.com/worldcoin/priv-orb-core/pull/2174) and [Oxide #1223](https://github.com/worldcoin/oxide/pull/1223); both pin the updated API.

## Validation

- Crate formatting, offline manifest metadata and whitespace checks pass.
- Nine unit tests use production functions, including empty plaintext and tampering with ciphertext or its tag. Known-vector, protobuf roundtrip and malformed-payload coverage remain.
- Tests and Clippy stop before compilation because the pinned relay dependency is not cached. They are not reported as passing. No dependencies or toolchains installed.

AI assistance: Codex assisted with the implementation, test call-site migration, and description.


ipcp-hpke/src/lib.rs
```diff
@@ -20,118 +19,183 @@ const KEY_LEN: usize = 32;
 const TAG_LEN: usize = 16;
 pub const PAYLOAD_OVERHEAD: usize = KEY_LEN + TAG_LEN;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct IpcpImageHpkePayload {
```
danielle-tfh: Can we not import the orb-relay-messages kind like qr-code package does? so we dont need to remake the type again?

ipcp-hpke/src/lib.rs
```diff
@@ -20,118 +19,183 @@ const KEY_LEN: usize = 32;
...
+    pub enc: Vec<u8>,
+    pub ciphertext: Vec<u8>,
+}
+
+pub struct PairingKey {
```
danielle-tfh, line 16: Do tese have to say Orb? Should they be generic for each party? Maybe PartyKey or something like that?

ipcp-hpke/src/lib.rs
```diff
@@ -20,118 +19,183 @@ const KEY_LEN: usize = 32;
...
+    orb_private_key: <Profile as Kem>::PrivateKey,
+    orb_public_key: [u8; KEY_LEN],
+}
+
+impl PairingKey {
```
danielle-tfh, line 27: iPCPPairingKey

ipcp-hpke/src/lib.rs
```diff
@@ -20,118 +19,183 @@ const KEY_LEN: usize = 32;
...
+    EncryptionTask(#[source] tokio::task::JoinError),
+    #[error("iPCP image decryption failed")]
     Decryption,
 }
 
-pub fn encrypt_ipcp(orb_public_key: &[u8], plaintext: &[u8]) -> Result<Vec<u8>, Error> {
-    encrypt_with_entropy(orb_public_key, plaintext, getrandom::fill)
+pub fn encrypt_ipcp_image_payload(
+    orb_public_key: Vec<u8>,
+    ipcp_image: Vec<u8>,
```
danielle-tfh: Why does this not use the apps private key?

Can this be part of a function for the PairingKeyAbove?

ipcp-hpke/src/lib.rs
```diff
@@ -20,118 +19,183 @@ const KEY_LEN: usize = 32;
...
+    orb_private_key: <Profile as Kem>::PrivateKey,
+    orb_public_key: [u8; KEY_LEN],
+}
+
+impl PairingKey {
+    pub fn new() -> Result<Self, Error> {
+        Self::with_randomness(getrandom::fill)
+    }
+
+    fn with_randomness(
```
danielle-tfh: Why does the library not check the entropy here?

Do we need to manually do it?

ipcp-hpke/src/tests.rs
```diff
@@ -0,0 +1,447 @@
...
+        );
+        assert!(matches!(
+            orb_pairing_key.decrypt_ipcp_image_payload(&malformed),
+            Err(Error::InvalidPayload)
+        ));
+    }
+}
+
+#[test]
+fn modification_at_every_payload_byte_is_rejected() {
```
danielle-tfh: can we cut the tests that only fail if hpke or x25519-dalek regress? `modification_at_every_payload_byte_is_rejected` is AES-GCM doing its job, `invalid_orb_public_key_lengths_are_rejected` is `from_bytes`, `pairing_key_matches_existing_x25519_fixture` is upstream keygen. the field-split and info/aad ones are what actually catch a change in here.

id keep `all_zero_and_low_order_public_keys_are_rejected` though, worth pinning across a dep bump.

ipcp-hpke/src/lib.rs
```diff
@@ -20,118 +19,183 @@ const KEY_LEN: usize = 32;
...
-    #[error("iPCP decryption failed")]
+    #[error("iPCP image encryption task failed")]
+    EncryptionTask(#[source] tokio::task::JoinError),
+    #[error("iPCP image decryption failed")]
     Decryption,
 }
 
-pub fn encrypt_ipcp(orb_public_key: &[u8], plaintext: &[u8]) -> Result<Vec<u8>, Error> {
-    encrypt_with_entropy(orb_public_key, plaintext, getrandom::fill)
+pub fn encrypt_ipcp_image_payload(
```
danielle-tfh: can we drop the tokio dep and make this sync? spawn_blocking is the callers decision, and taking it here pulls tokio into a crypto leaf that orb-core and oxide both consume. aes-gcm over one image is single digit ms anyway, and a caller that does want it off the runtime can do it in one line at the call site.

that also drops the `impl Future` return and `Error::EncryptionTask`.

ipcp-hpke/src/lib.rs
melmols: Keeping fallible OS randomness handling: HPKE’s built-in path panics on RNG failure; ours returns `Error::Randomness`. This checks availability, not entropy quality.

ipcp-hpke/src/lib.rs
```diff
@@ -8,130 +8,177 @@ use hpke::{
...
-    info: &[u8],
-    aad: &[u8],
-    entropy: &mut EphemeralEntropy,
-) -> Result<Vec<u8>, Error> {
-    let public_key = <Profile as Kem>::PublicKey::from_bytes(orb_public_key)
+    fn with_randomness(
+        fill: impl FnOnce(&mut [u8]) -> Result<(), getrandom::Error>,
+    ) -> Result<Self, Error> {
+        let mut pairing_private_key_bytes = Zeroizing::new([0; KEY_LEN]);
+        fill(pairing_private_key_bytes.as_mut()).map_err(|_| Error::Randomness)?;
```
danielle-tfh: Can we enable hpke’s `getrandom` feature and use `Profile::gen_keypair()` here (and `single_shot_seal_inout_detached()` below)? A system RNG failure means we cannot create a safe key and have no fallback, so the crate’s fail-stop behavior is right.

danielle-tfh (approved):

ipcp-hpke/src/lib.rs
```diff
@@ -1,425 +1,113 @@
...
-    fn bytes(fixture: &Value, field: &str) -> Vec<u8> {
-        hex::decode(fixture[field].as_str().unwrap()).unwrap()
-    }
-
-    fn entropy(fixture: &Value) -> EphemeralEntropy {
-        EphemeralEntropy {
-            bytes: Zeroizing::new(bytes(fixture, "ikmE").try_into().unwrap()),
-            position: 0,
+impl PairingKey {
+    pub fn new_pairing_key() -> Self {
```
andronat:
```suggestion
    pub fn new() -> Self {
```

ipcp-hpke/src/lib.rs
```diff
@@ -1,425 +1,113 @@
...
-        output.copy_from_slice(
-            self.bytes
-                .get(self.position..end)
-                .expect("X25519 entropy exhausted"),
-        );
-        self.position = end;
-        Ok(())
-    }
+pub struct PairingKey {
+    pairing_private_key: <Profile as Kem>::PrivateKey,
```
andronat:
```suggestion
    private_key: <Profile as Kem>::PrivateKey,
```

ipcp-hpke/src/lib.rs
```diff
@@ -1,425 +1,113 @@
...
+    pub pairing_public_key: [u8; KEY_LEN],
```
andronat:
```suggestion
    pub public_key: [u8; KEY_LEN],
```

ipcp-hpke/src/lib.rs
```diff
@@ -1,425 +1,113 @@
...
-                &bytes(&f, "pt"),
+        .map_err(|_| Error::Decryption)?;
+        Ok(plaintext)
+    }
+
+    pub fn encrypt(
+        recipient_public_key: &[u8],
+        plaintext: Vec<u8>,
+        info: &[u8],
+        aad: &[u8],
```
andronat: as discussed let's keep it minimal and not expose complexity of choice to iPCP consumers. Let's decide what these values are internally in this lib.

ipcp-hpke/src/lib.rs
```diff
@@ -1,425 +1,109 @@
...
-                info,
-                aad,
-                &mut entropy(&f),
+        .map_err(|_| Error::Decryption)?;
+        Ok(plaintext)
+    }
+
+    pub fn encrypt(
+        recipient_public_key: &[u8],
+        plaintext: Vec<u8>,
```
andronat: as we discussed offline we you can wrap this type in zeroed you can force the user to take a better practice of putting their data in zeroing lib.

danielle-tfh (approved):

vmenge (approved):

PR #1417, opened 2026-09-19
refactor(ipcp): add a validated recipient public key type

andronat:
Introduce an opaque RecipientPublicKey with fallible byte parsing and serialization, and require it in PairingKey::encrypt. Existing HPKE keys can be wrapped directly with from_pk without serializing and reparsing.

Cover malformed key lengths and public key serialization, and update encryption tests to exercise the typed API.


danielle-tfh (approved):

ipcp-hpke/src/lib.rs
```diff
@@ -74,6 +74,28 @@ impl PairingKey {
     }
 }
 
+pub struct RecipientPublicKey(<Profile as Kem>::PublicKey);
```
danielle-tfh, line 77: nice!

danielle-tfh (approved):

danielle-tfh (approved):

